PR #3, opened 2017-02-10
Index page review except select

reubenvarghese1:
This PR is now ready to be merged after review and contains the set of changes for the Index page


src/main/ui/app/component/AddOnList.jsx
```diff
@@ -1,21 +1,43 @@
...
+            )
         }
         else {
             return (
-                    <div>
+                   <div>
+		   <Table className="maintable">
+    			<thead>
+      			<tr>
+			<th><i className='fa fa-2x fa-globe' aria-hidden="true"></i></th>
```
djazayeri: Git rid of the globe icon here. Better for it to be blank, since we don't want to draw the user's eye here.

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,42 @@
...
-                            </h4>
-                            {addon.description}
-                        </div>
-                    </div>
-                </Link>
+                     	</Media.Left>
+                   <Media.Body> 
+                            
+    			
+		            <Accordion>
```
djazayeri: I don't like the accordion effect here. Just put the description text directly. (Maybe later we'll need to do something to shorten the text if it's too long.)

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,42 @@
...
+    <Panel header={title} className="listgrp" eventKey="1">
+      {addon.description} 
+<p><div className="takeright"><Button className="spacing" href={link} bsSize="small">More Info</Button>
+<Button  href={link} bsStyle="primary" bsSize="small">Download</Button></div>
```
djazayeri: Let's not have a direct download button. They should need to click on the page to get more info first.
(In the future I have in mind that on the show page we'll do some additional logic to determine which is the right version of the module for their OpenMRS Version)

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,42 @@
...
-                    </div>
-                </Link>
+                     	</Media.Left>
+                   <Media.Body> 
+                            
+    			
+		            <Accordion>
+    <Panel header={title} className="listgrp" eventKey="1">
+      {addon.description} 
+<p><div className="takeright"><Button className="spacing" href={link} bsSize="small">More Info</Button>
```
djazayeri: I think I prefer the original effect where there's a Link around the whole item, so clicking anywhere opens the Show view. (Instead of a More Info button.)

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,42 @@
...
 import {Link} from "react-router";
+import { ListGroupItem , Badge, Accordion , Panel , Media , Button } from 'react-bootstrap';
 
 export default class AddOn extends Component {
 
     render() {
         const addon = this.props.addon;
         const icon = addon.type === 'OWA' ? 'fa fa-2x fa-globe' : 'fa fa-2x fa-puzzle-piece';
-        const link = this.props.version ? `/show/${addon.uid}?highlightVersion=${this.props.version}` : `/show/${addon.uid}`;
+        const link = this.props.version ? `#/show/${addon.uid}?highlightVersion=${this.props.version}` : `#/show/${addon.uid}`;
```
djazayeri: I'm surprised this is necessary. What changed to require #?

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -57,31 +57,40 @@ export default class AddOnSearch extends Component {
...
-                        <option value="OWA">Open Web App (OWA)</option>
-                    </select>
-
-                    <br/>
+                <div className="row col-md-12 col-sm-12 col-xs-12 pushdown">
+                    <div>
+    <div className="input-group">
+    
+      <div className="input-group-btn">
+        <button type="button" className="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Type <span className="caret"></span></button>
```
djazayeri: I think you need `<DropdownButton>` for this: https://react-bootstrap.github.io/components.html#btn-dropdowns

src/main/ui/app/route/About.jsx
```diff
@@ -9,7 +9,7 @@ export default class About extends Component {
                     <h3>About</h3>
 
                     <p>
-                        <Link to="/indexingStatus">Indexing Status</Link>
+                        <Link className="btn btn-lg btn-primary col-md-1 col-sm-12 col-xs-12 buttonfix" to="/indexingStatus">Indexing Status</Link>
```
djazayeri: Don't make this a button, just leave it as a plain link. This is advanced functionality that's kind of only there for admin troubleshooting, so I don't want to draw clicks to it.

src/main/ui/app/route/App.jsx
```diff
@@ -5,7 +5,7 @@ export default class App extends Component {
 
     render() {
         return (
-                <div className="container-fluid">
```
djazayeri: I vote to leave this fluid.

src/main/ui/app/sass/addonindex.scss
```diff
@@ -26,14 +26,32 @@
   float: right;
 
 }
+.maintable{
+margin-bottom:0;
+}
+
+
+.titletype{
```
djazayeri: Can't we use a standard bootstrap class for this stuff? I'm surprised we need to this much customization.

src/main/ui/app/sass/addonindex.scss
```diff
@@ -26,14 +26,32 @@
...
+
+}
+.mainlist{
+border-width: 0;
+}
 .pushdown {
   margin-top: 40px;
 }
 
 .textdec {
```
djazayeri: What is textdec supposed to mean, anyway? It's not an abbreviation I'm familiar with. In general I'd prefer css class names to be dash-separated, e.g. "main-list" instead of "mainlist"

src/main/ui/app/sass/addonindex.scss
```diff
@@ -65,15 +83,34 @@ tr {
...
   margin-top: 3em;
 }
+a:hover, a:visited, a:link, a:active
+{
+    text-decoration: none;
+}
+.listgrp{
+padding : 0;
+}
+.takeright{
```
djazayeri: Why can't we use boostrap's built-in "pull-right" for this?

src/main/ui/index.ejs
```diff
@@ -1,13 +1,19 @@
...
+         <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.min.css">
+
     <script src="https://unpkg.com/react@15/dist/react.min.js"></script>
     <script src="https://unpkg.com/react-dom@15/dist/react-dom.min.js"></script>
     <script src="https://unpkg.com/react-router/umd/ReactRouter.min.js"></script>
     <script src="https://use.fontawesome.com/75c2bbc0c8.js"></script>
 </head>
 <body>
 <div id="root"></div>
+<footer>
```
djazayeri: I vote to remove the footer. It doesn't add useful info for the user.

src/main/ui/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "isomorphic-fetch": "^2.2.1",
     "react": "^15.4.2",
+    "react-bootstrap": "^0.30.7",
```
djazayeri: Eventually we should include this from CDN (which would be a change to index.ejs and to webpack.config.js), but we don't need to just yet.

src/main/ui/app/component/AddOn.jsx
reubenvarghese1: The initial path was index/#/(Not sure why) and hence without this change none of the other pages would load,hence I had to add this

src/main/ui/app/route/About.jsx
reubenvarghese1: Will it be part of the main app?

src/main/ui/app/sass/addonindex.scss
reubenvarghese1: Minor repositioning of the heading is needed as it is not part of the list unlike a table

src/main/ui/app/sass/addonindex.scss
reubenvarghese1: Textdec is the class used for styling the error codes displayed (I know the name doesn't make sense )

src/main/ui/app/sass/addonindex.scss
reubenvarghese1: Was not aware of the existence of such a class. Will change

src/main/ui/app/route/About.jsx
djazayeri: We might even hide it (e.g. it's there only if you know the URL). For now leave it as is, but don't make it more obvious.

src/main/ui/app/component/AddOn.jsx
reubenvarghese1: Just realized that it is not required if we use {Link} from "react-router" . This adds the # by default

src/main/ui/app/sass/addonindex.scss
djazayeri: Did you introduce that name or is it a standard bootstrap thing?

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,39 @@
...
 
 export default class AddOn extends Component {
 
     render() {
         const addon = this.props.addon;
         const icon = addon.type === 'OWA' ? 'fa fa-2x fa-globe' : 'fa fa-2x fa-puzzle-piece';
         const link = this.props.version ? `/show/${addon.uid}?highlightVersion=${this.props.version}` : `/show/${addon.uid}`;
+	
+	const title = (
+ <div>{addon.name} <div className="pull-right">{addon.type}</div></div>
```
djazayeri: Should format this like in the other files

src/main/ui/app/component/AddOn.jsx
```diff
@@ -1,30 +1,39 @@
...
-                            </h4>
-                            {addon.description}
-                        </div>
-                    </div>
+                     	</Media.Left>
+                   <Media.Body> 
+                            
+    			
+		            
+    <Panel header={title} className="list-grp" eventKey="1">
```
djazayeri: Formatting (clean up the extra blank lines, and also indent it like the other things)

djazayeri: No need for `eventKey="1"` here, right?

src/main/ui/app/component/AddOnList.jsx
```diff
@@ -1,21 +1,43 @@
...
+      </section>
+
+            )
         }
         else {
             return (
-                    <div>
+                   <div>
+		   <Table className="maintable">
+    			<thead>
```
djazayeri: Looking at the actual UI, I don't see a need for this header. It's pretty obvious what things are, and this takes up valuable space toward the top of the screen.

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -56,32 +56,45 @@ export default class AddOnSearch extends Component {
...
-                    Type:
-                    <select onChange={event => this.setType(event.target.value)}>
-                        <option value="">All Types</option>
-                        <option value="OMOD">Module (OMOD)</option>
-                        <option value="OWA">Open Web App (OWA)</option>
-                    </select>
 
-                    <br/>
+const title = (
+<span>Type: {this.state.type}</span>
```
djazayeri: Actually we should change this so it's displaying as one of these strings:

* All Types
* Modules
* Open Web Apps

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -56,32 +56,45 @@ export default class AddOnSearch extends Component {
...
+);
+        return (
+                <div className="row col-md-12 col-sm-12 col-xs-12 pushdown">
+                    <div>
+    <div className="input-group">
```
djazayeri: Indentation throughout here is distracting.
Also, having inconsistent use of blank lines is distracting, e.g. the div around DropdownButton has a blank line at the start but not at the end.

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -56,32 +56,45 @@ export default class AddOnSearch extends Component {
...
+        <MenuItem eventKey="OMOD" >Module (OMOD)</MenuItem>
+        <MenuItem eventKey="OWA">Open Web App (OWA)</MenuItem>
+      </DropdownButton>
+      </div>
+  
+  
                     <DebounceInput
                             placeholder="Search..."
-                            className="search"
+                            className="form-control col-md-10 col-sm-10 col-xs-10"
```
djazayeri: As a general comment, I thought that with bootstrap's grid system, if you're going to set the column size to be the same on different device sizes, you don't need to duplicate it. E.g. just doing `col-md-10` is enough without repeating it for sm and xs. (But I'm not 100% sure about that.)

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -56,32 +56,45 @@ export default class AddOnSearch extends Component {
...
                             minLength={1}
                             debounceTimeout={500}
                             onChange={event => this.setQuery(event.target.value)}/>
-                    <i className="fa fa-search fa-2x"></i>
-
-                    <br/>
+
```
djazayeri: clean up blank lines :-)

djazayeri:
I merged this using "Squash and Merge" so that it will appear as only a single commit in the main codebase. (A bit about that is [written here](https://github.com/blog/2141-squash-your-commits), in case you're not familiar with the idea.)

You  may need to fork the repository again. Or maybe git will automagically fix this when you do a pull from upstream/master. (I've never used this workflow in github before.)

src/main/ui/app/sass/addonindex.scss
reubenvarghese1: I introduced it .I'll change it to something more sensible

src/main/ui/app/component/AddOnSearch.jsx
reubenvarghese1: Sorry for my bad programming practices ,will format the code in idea

src/main/ui/app/component/AddOnSearch.jsx
reubenvarghese1: I reckon you are correct :+1:

src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -56,32 +56,45 @@ export default class AddOnSearch extends Component {
...
+  
+  
                     <DebounceInput
                             placeholder="Search..."
-                            className="search"
+                            className="form-control col-md-10 col-sm-10 col-xs-10"
                             minLength={1}
                             debounceTimeout={500}
                             onChange={event => this.setQuery(event.target.value)}/>
-                    <i className="fa fa-search fa-2x"></i>
```
reubenvarghese1: I was wondering if we require the puzzle icon next to the module name as we already have the type specified

djazayeri: Oh, good point, I noticed this also. As it stands now, there's no value to having the icons, so I think we could just remove them.

In the future I'd like to let each addon specify its own icon (in the add-ons-to-index.json file).

So, yes, for now it's fine to remove the icons.

src/main/ui/app/component/AddOnSearch.jsx
reubenvarghese1: I think we were wrong.I just came across a situation which confirmed that we have to explicitly give the values for all three sizes or else the condition did not apply

PR #4, opened 2017-02-15
Styled the modules page 

reubenvarghese1:
Styled the modules page and a few other minor changes including certain changes that have been implemented based on previous pr comments


src/main/ui/app/component/AddOnSearch.jsx
```diff
@@ -2,7 +2,7 @@ import {Component} from "react";
 import fetch from "isomorphic-fetch";
 import AddOnList from "./AddOnList";
 import DebounceInput from "react-debounce-input";
-import { DropdownButton , MenuItem } from 'react-bootstrap';
+import { DropdownButton , MenuItem , Glyphicon} from 'react-bootstrap';
```
djazayeri: Is this added for a reason?

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
+                               );
+
+            const versionmeaning = (
+                                <Tooltip id="tooltip"><strong>{addon.name} version</strong></Tooltip>
+                            );
+	    const versioncountmeaning = (
+                                <Tooltip id="tooltip"><strong>No. of builds of {addon.name}</strong></Tooltip>
+                            );
+            const requirementmeaning = (
+                                <Tooltip id="tooltip"><strong>This shows the minimum version of OpenMRS required to be able to run this module</strong></Tooltip>
```
djazayeri: rephrase as ...minimum version of the OpenMRS Platform required...

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
-                            <span className={statusClass}>{addon.status}</span>
-                        </h1>
-                        <h3>{addon.description}</h3>
-                        {hosted}
-                        <p>
-                            Maintained by:
-                            {addon.maintainers.map(m => {
+
+                    <div className="showpage-body">
+<Panel header={showtitle}>
```
djazayeri: formatting :-)

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
+{check}
+<tr>
+        <th>Maintained by: </th>
+        <td><h5>{addon.maintainers.map(m => {
                                 if (m.url) {
                                     return (
-                                            <a target="_blank" href={m.url}>{m.name}</a>
-                                    )
+                                    
+                                   <Button bsSize="small" href={m.url}>{m.name}</Button>                                   )
```
djazayeri: I haven't looked at how this looks in the UI yet, but why would we display this as a button?

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
                                 }
                             })}
-                        </p>
-                        Versions:
-                        <ul>
-                            {addon.versions.map(v => {
+    </h5></td>
+</tr>
+<tr>
+        <th>Version Count  <OverlayTrigger placement="right" overlay={versioncountmeaning}>
```
djazayeri: At first thought, if the meaning of this is so unclear that we need to provide a tooltip, I would prefer to just make the text clearer.

src/main/ui/app/component/AddOnSearch.jsx
reubenvarghese1: I was going to add a search icon to it but then it didn't look all that good so I removed it .The gist is that this is no longer needed and I'll revert that change ;)

src/main/ui/app/route/Show.jsx
reubenvarghese1: I felt it was looking pretty plain if it was just the names.  We could possibly get the images of the devs from their profile just like in modulus.In the meantime I'll think of other alternatives to make it look better without using a button

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
+<Table condensed hover>
+    <thead>
+      <tr>
+        <th className="col-md-2 col-sm-2 col-xs-2">Version<OverlayTrigger placement="right" overlay={versionmeaning}>
+      <Glyphicon glyph="glyphicon glyphicon-question-sign" />
+    </OverlayTrigger></th>
+        <th className="col-md-6 col-sm-6 col-xs-6">OpenMRS Version Requirement<OverlayTrigger placement="right" overlay={requirementmeaning}>
+      <Glyphicon glyph="glyphicon glyphicon-question-sign" />
+    </OverlayTrigger></th>
+        <th className="col-md-2 col-sm-2 col-xs-2">Not sure</th>
```
reubenvarghese1: I wasn't able to figure out the info it provides.It's only output is 1,2,3 or something like that so shall I remove that column?

src/main/ui/app/route/Show.jsx
```diff
@@ -39,61 +39,128 @@ export default class Show extends Component {
...
-                        <h3>{addon.description}</h3>
-                        {hosted}
-                        <p>
-                            Maintained by:
-                            {addon.maintainers.map(m => {
+
+                    <div className="showpage-body">
+<Panel header={showtitle}>
+<h4 className="lead">{addon.description}</h4>
+<div className="col-md-6 col-sm-12 col-xs-12 left-margin">
```
reubenvarghese1: I felt it was too wide on a normal page considering the amount of info that it provides.As in it looks way too empty since the actual data only takes about half of the page

src/main/ui/app/route/Show.jsx
djazayeri: So, as I said in a later comment, we shouldn't decorate things just to make them look "good" but we should apply visual effects to draw the eye more (or less). In this case we don't actually want to draw the eye here, so we should just leave these as plain links.

src/main/ui/app/route/Show.jsx
djazayeri: Is there a way to make the first column be fixed-width? When I'm looking at http://localhost:8090/#/show/org.openmrs.module.formentry the maintainers break onto two lines.

djazayeri: Alternately, we should consider not using a table for this.

djazayeri:
I made a few changes to this, and manually merged it.
(There was a missing import in Show.jsx. I fixed the formatting. And I did a bit more UI rework.)

PR #5, opened 2017-03-13
Changed resource link for react router

reubenvarghese1:
The older link caused the site to be unable to fetch the react router.js file. This commit basically updates the old broken link.


reubenvarghese1:
@djazayeri  I just saw that the live site was not loading and here is a fix for the same.

PR #10, opened 2017-05-16
Named list made full screen

reubenvarghese1:


dkayiwa:
@djazayeri  are you happy with this? 😊

src/main/ui/app/component/NamedList.jsx
```diff
@@ -27,9 +27,8 @@ export default class NamedList extends Component {
         const COLUMNS_IF_SIZE = 5;
 
         const list = this.props.list;
-        const colSize = list.addOns.length > COLUMNS_IF_SIZE ?
```
djazayeri: Ideally we'd preserve the behavior of using two columns to show more items on larger screens.

So, first I'd like to try bringing back that behavior. (But if we're going to remove it, then we should also move the COLUMNS_IF_SIZE const above.

djazayeri: @reubenvarghese1 okay, so I went back and recreated my old code for this and I posted here to ask people which works better: https://talk.openmrs.org/t/replacing-modulus-design-thoughts/9619/37

Looking at them side-by-size I now think that the 2-column view is misguided, since the eye has to move horizontally as well as vertically, and this is distracting. So, unless several people reply to my post saying they like the 2-column view, we should go with your solution. (And if we do this, also remove the COLUMNS_IF_SIZE constant, and also the half function earlier in this file.)

If we do end up going with the 2-column version then the code should look like

    render() {
        const COLUMNS_IF_SIZE = 10;

        const list = this.props.list;
        const colSize = list.addOns.length >= COLUMNS_IF_SIZE ?
                        "col-md-6 col-sm-12 col-xs-12" :
                        "col-md-12 col-sm-12 col-xs-12";

        return (
                <div>
                    <h1>{list.name}</h1>
                    <h3>{list.description}</h3>
                    { list.addOns.length >= COLUMNS_IF_SIZE ?
                      <div className="row">
                          <div className={colSize}>
                              { this.half(list.addOns, 0).map(addon => this.display(addon)) }
                          </div>
                          <div className={colSize}>
                              { this.half(list.addOns, 1).map(addon => this.display(addon)) }
                          </div>
                      </div>
                      :
                      <div className="row">
                          <div className={colSize}>
                              { list.addOns.map(addon => this.display(addon)) }
                          </div>
                      </div>
                    }
                </div>
        )
    }

reubenvarghese1: @djazayeri I'd say we could wait for a reply. If we don't get one then we'll go with removing the feature . Honestly, looking at the screenshots the 2-column version looks good as well but the only thing is as to whether the output is worth the effort. If we go with the 2-sided version here then there are more places such as list of downloads , search results etc which could use this as well.

reubenvarghese1:
I am closing this PR temporarily so that I may open another PR to fix another issue

PR #12, opened 2017-05-17
Made named list full screen

reubenvarghese1:
Named list is now full screen based on community feedback (Currently still to be decided but in case we choose this route then we can merge this PR)

### DIFF
--- a/src/main/ui/app/component/NamedList.jsx
+++ b/src/main/ui/app/component/NamedList.jsx
@@ -17,7 +17,6 @@ export default class NamedList extends Component {
     }
 
     render() {
-        const COLUMNS_IF_SIZE = 5;
 
         const list = this.props.list;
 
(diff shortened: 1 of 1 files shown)


src/main/ui/app/component/NamedList.jsx
```diff
@@ -27,16 +27,13 @@ export default class NamedList extends Component {
         const COLUMNS_IF_SIZE = 5;
```
djazayeri: Please also delete this, as well as the "half" function above. These are no longer used, right?

reubenvarghese1: Sorry , don't know why I forgot about that. I've deleted that as well .

djazayeri: I meant also delete the line `const COLUMNS_IF_SIZE = 5;`

reubenvarghese1: Spare me if I still have missed something. Haven't really had much time due to the exams 😉

PR #16, opened 2017-06-09
AO 2 - Improve current search algorithm

reubenvarghese1:
The search algorithm has been improved to a great extent although we can expect an even more improved version of this same algorithm with some future PRs


src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -100,16 +100,21 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
-			boolQB.should(QueryBuilders.matchQuery("name", query).boost(4.0f));
-			boolQB.should(QueryBuilders.prefixQuery("_all", query).boost(1.5f));
-			boolQB.should(QueryBuilders.matchPhrasePrefixQuery("_all", query).slop(2).fuzziness("AUTO"));
+			//Following clause return a module whose UID matches exactly
+			boolQB.should(QueryBuilders.termQuery("_id", query).boost(1500.0f));
+			boolQB.should(QueryBuilders.prefixQuery("name", query).boost(4.0f));
+			boolQB.should(QueryBuilders.termQuery("tags", query).boost(1500.0f));
+			boolQB.should(QueryBuilders.matchQuery("name", query).boost(2.0f));
+			boolQB.should(QueryBuilders.matchQuery("description", query).boost(0.5f));
+			//boolQB.should(QueryBuilders.prefixQuery("name", query).boost(2.5f));
```
djazayeri: Don't leave unused commented code.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -100,16 +100,21 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
 			boolQB.filter(QueryBuilders.matchQuery("type", type));
 		}
 		if (query != null) {
-			boolQB.should(QueryBuilders.matchQuery("name", query).boost(4.0f));
-			boolQB.should(QueryBuilders.prefixQuery("_all", query).boost(1.5f));
-			boolQB.should(QueryBuilders.matchPhrasePrefixQuery("_all", query).slop(2).fuzziness("AUTO"));
+			//Following clause return a module whose UID matches exactly
```
djazayeri: I suggest you have an equivalent comment for each one of these lines, describing what it does in plain English.

And, add some newlines, so it ends up like

```
// exact id match scores the highest
boolQB.should(...)

// next comment
boolQB.should(...)
```

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -100,16 +100,21 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
+			boolQB.should(QueryBuilders.termQuery("_id", query).boost(1500.0f));
+			boolQB.should(QueryBuilders.prefixQuery("name", query).boost(4.0f));
+			boolQB.should(QueryBuilders.termQuery("tags", query).boost(1500.0f));
```
djazayeri: Boost this slightly less, e.g. so we're sure that an exact id match scores a bit higher than an exact tag match.

src/main/resources/elasticsearch/addOnInfoAndVersions-mappings.json
```diff
@@ -14,7 +14,8 @@
         "type": "text"
       },
       "tags": {
-        "type": "keyword"
+        "type": "keyword",
+	"index": "not_analyzed"
```
djazayeri: Is this necessary?

https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html says:
> Keyword fields are only searchable by their exact value.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
djazayeri: Also, sort these by boost, so that someone reading the code can immediately see the most important clauses first.

reubenvarghese1: When an exact uid matches then it never shows any other result. Similarly when a tag matches it's never possible for a UID to match and hence we do not have to worry about them clashing. Either way since you have mentioned it, I shall change it ☺️

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: Sorry, my bad

src/main/resources/elasticsearch/addOnInfoAndVersions-mappings.json
reubenvarghese1: Yes , this is very important . I did read what you pointed out but for some reason if I remember right then without explicitly denoting it as non analysed , it was still getting tokenized and hence not matching perfectly. Actually the - in "form-entry" acts as a delimiter and it gets tokenized. I'll try changing it and get back.

reubenvarghese1: Just check and turns out I was wrong. Don't know why it happened before but now just the type keyword is sufficient . I have changed it

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -97,19 +97,36 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
 		
 		BoolQueryBuilder boolQB = QueryBuilders.boolQuery();
 		if (type != null) {
+			//Gets all modules of the type specified by variable type
```
djazayeri: As a general style guide, you should never have a code comment that does not provide extra information to the reader.
This comment is bad for two reasons:

1. don't say "by variable type" because it's very obvious to someone looking at the code that type is a variable, and because of its name it's self-documenting
2. the whole line doesn't provide any additional info that you couldn't get from looking at the one line of code.

The thing that _would_ be useful to say in this comment is to provide a bit of information to someone who may be reading this, but is unfamiliar with the ES query API. E.g. most people won't know specifically what a "match" query is. So I would have written as the comment here:

    // exact match on type

This would provide the reader with the most information about what this does, while taking the least time to read.

Also, as I type this, I don't see why we didn't use a `term` query here instead of a `match`.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -97,19 +97,36 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
 			boolQB.filter(QueryBuilders.matchQuery("type", type));
 		}
 		if (query != null) {
-			boolQB.should(QueryBuilders.matchQuery("name", query).boost(4.0f));
-			boolQB.should(QueryBuilders.prefixQuery("_all", query).boost(1.5f));
-			boolQB.should(QueryBuilders.matchPhrasePrefixQuery("_all", query).slop(2).fuzziness("AUTO"));
+			//Following clause returns a module whose UID matches exactly(Highest priority)
```
djazayeri: Like I said in the comment above this one, you should make the comment shorter and more efficient.

* Don't say "Following clause" because it's obvious from the code spacing that this comment applies to the next line
* Don't say "returns all modules" because anyone reading this block of code knows that we're building a query, and hence it's going to query for a group of addons.

I would have written this as

    // Top priority: exact match on _id

(And, apply this same comment to the next 4 comments also.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -97,19 +97,36 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
+			//Following clause returns all modules whose prefix of name matches query(Medium priority)
+			boolQB.should(QueryBuilders.prefixQuery("name", query).boost(4.0f));
+
+			//Following clause returns all modules whose name matches the query(Medium priority)
+			boolQB.should(QueryBuilders.matchQuery("name", query).boost(2.0f));
+
+			//Following clause returns all modules whose description words match the query(Low priority)
+			boolQB.should(QueryBuilders.matchQuery("description", query).boost(0.5f));
+
+			//Allow for spelling mistake while searching for a particular module
```
djazayeri: This comment is good.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: I understand what you mean. I'll commit the changes with the modified comments and hopefully learn from what you've said:). And to address the usgae of match vs term here , I do agree that term would've been the right choice but it doesn't actually make a difference as type is just word and hence there's nothing to be tokenized.

PR #17, opened 2017-06-13
AO-3 Improved Tagging support of modules

reubenvarghese1:
Improved Tagging support of modules:

Includes creation of

1. Tagging API
2. Deleted obsolete InMemoryIndex files 
3. Added link between API and UI


src/main/java/org/openmrs/addonindex/service/Index.java
```diff
@@ -27,7 +27,7 @@
 	
 	Collection<AddOnInfoAndVersions> getAllByType(AddOnType type) throws Exception;
 	
-	Collection<AddOnInfoAndVersions> getByTag(String tag) throws Exception;
+	Collection<AddOnInfoSummary> getByTag(String tag) throws Exception;
```
djazayeri: Why are you changing this to return less info?

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
```diff
@@ -40,6 +40,11 @@ public AddOnController(Index index) {
 	                                           @RequestParam(value = "q", required = false) String query) throws Exception {
 		return index.search(type, query);
 	}
+
+	@RequestMapping(method = RequestMethod.GET, value = "/api/v1/tags")
```
djazayeri: This is the wrong URL for this. It implies a `tag` resource, i.e. you're fetching tags, not the addons. Two options:

1. this would be at "/api/v1/addon?tag=Tag"
2. this would be at "/api/v1/tag/{tagName}/"

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -156,14 +156,14 @@ public AddOnInfoAndVersions getByUid(String uid) throws IOException {
...
-	public Collection<AddOnInfoAndVersions> getByTag(String tag) throws Exception {
+	public Collection<AddOnInfoSummary> getByTag(String tag) throws Exception {
 		SearchResult result = client.execute(new Search.Builder(new SearchSourceBuilder()
 				.size(SEARCH_SIZE)
 				.query(QueryBuilders.matchQuery("tags", tag)).toString())
 				.addIndex(AddOnInfoAndVersions.ES_INDEX)
 				.build());
 		return result.getHits(AddOnInfoAndVersions.class).stream()
-				.map(sr -> sr.source)
+				.map(sr -> new AddOnInfoSummary(sr.source))
```
djazayeri: I don't see why you'd change the code here to discard information.

src/main/ui/app/route/TagPage.jsx
```diff
@@ -0,0 +1,93 @@
...
+    componentDidUpdate(prevProps) {
+        if (this.props.location.query !== prevProps.location.query) {
+            this.getTag();
+        }
+    }
+
+    getTag() {
+        const tag = this.props.location.query.tag;
+    
+        console.log(tag);
```
djazayeri: remove console.log statements before sending the PR :-)

src/main/ui/app/route/TagPage.jsx
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+import {Component} from "react";
+import fetch from "isomorphic-fetch";
+import AddOnList from "../component/AddOnList";
+
+export default class TagPage extends Component {
```
djazayeri: Looking at this, I wonder if we'd have less code duplication if we added this functionality to the existing search page...

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
reubenvarghese1: I was initially planning to do that but then how do I call 2 functions under the same mapping or rather can we?

src/main/ui/app/route/TagPage.jsx
reubenvarghese1: I didn't add it to the existing search page only due to one reason. Even there is a perfect tag match, the search results display more modules than the modules that actually have the tag.
It may be noted that the highest rank results are always the ones with the matching tag but it still returns more.
Example : https://addons-stg.openmrs.org/#/search?&q=form-entry 
Actual modules with the tag are only the Top 4 but it returns more than 10 results

src/main/ui/app/route/TagPage.jsx
reubenvarghese1: Sure, No problem :)

src/main/java/org/openmrs/addonindex/service/Index.java
reubenvarghese1: Cause I felt it returns more info than needed for our current use case. Let me know what you feel and accordingly I'll change it .

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: Same reason as above, let me know if you want me to change it back

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
djazayeri: In fact I would just handle this as a third optional parameter in the search method in this controller. And you'd add this as a third argument to the index.search method.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
djazayeri: Yes, just leave this as it was. There's no reason to discard information pre-emptively on the back-end. IIRC the purpose of the AddOnInfoAndSummary class is so we can send less information across the network to the REST client, but when we're on the server we want to keep all the info.

src/main/ui/app/route/TagPage.jsx
reubenvarghese1: Any thoughts about this one?

src/main/ui/app/route/TagPage.jsx
djazayeri: Sorry, in my head this was implied by one of my other comments, but I should have been explicit!

To correctly implement this feature, we should support `https://addons-stg.openmrs.org/#/search?&tag=form-entry`

Basically we should change the underlying search functionality so that it can accept 3 different optional parameters: q, type, tag.

The REST endpoint should also accept all three of these, and pass them down to the Java API.

And I assume that would allow us to use the same search page to handle everything without duplicating.

(But if that doesn't work for some reason it's okay to do what you're doing here; I haven't closely reviewed this file, but I will do so. I assume it's a lot of copy-paste from the List page?)

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -93,27 +93,39 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
 			boolQB.filter(QueryBuilders.matchQuery("type", type));
 		}
+		if (tag != null) {
+			
+		SearchResult result = client.execute(new Search.Builder(new SearchSourceBuilder()
+				.size(SEARCH_SIZE)
+				.query(QueryBuilders.matchQuery("tags", tag)).toString())
+				.addIndex(AddOnInfoAndVersions.ES_INDEX)
+				.build());
+		return result.getHits(AddOnInfoAndVersions.class).stream()
```
djazayeri: So, the logic you've implemented here means that it will _only_ search by tag. And this would be confusing for someone who wants to call this method, if you're allowed to search for type and query optionally together, or tag alone.

But I think it's trivial to support searching by whatever combination of criteria the caller passes in. You'd just add to boolQB like is done in the if/then block about 'type'

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -93,27 +93,39 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
...
+				.map(sr -> new AddOnInfoSummary(sr.source))
+				.collect(Collectors.toList());
+
```
djazayeri: Also, please try to keep the formatting clean. E.g. there should be no extraneous blank lines. (It's fine to have an intentional blank line to make the code more readable, but that should almost never be the first or last line of a block.

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -33,7 +33,8 @@ export default class SearchPage extends Component {
     doSearch() {
         const query = this.props.location.query.q;
         const type = this.props.location.query.type;
-        const searchKey = `type:${type ? type : "all"} query:${query ? query : ""}`;
+	const tag = this.props.location.query.tag;
```
djazayeri: Formatting; this should have the same indentation as the lines before and after.

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -33,7 +33,8 @@ export default class SearchPage extends Component {
...
+        const searchKey = `type:${type ? type : "all"} query:${query ? query : ""} tag:${tag ? tag : ""}`;
 
         console.log(searchKey);
```
djazayeri: get rid of these. (were they leftover from my code?)

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -85,7 +89,7 @@ export default class SearchPage extends Component {
                         <SearchBox initialQuery={this.props.location.query.q}/>
 
                         {this.state.latestSearch ?
-                         <div>Searching for {this.state.latestSearch}</div>
+                         <div>Searching for {this.state.latestSearch} and </div>
```
djazayeri: Why add the word "and" here?

src/main/ui/app/route/Show.jsx
```diff
@@ -74,7 +76,7 @@ export default class Show extends Component {
             <Label bsStyle={statusClass}>{status}</Label>
             {addon.tags ?
              addon.tags.map(t =>
-                 <Label bsStyle="default">{t}</Label>
+                 <Link to={`/search?&tag=${t}`}><Label bsStyle="default">{t}</Label></Link>
```
djazayeri: With this a user can find one module, and then browse to other modules with the same tag. But we also want to give them a page that lists all the tags and lets them click on one. (Fine to do that as part of a different ticket, so we don't get further derailed from our originally scheduled work this week.)

src/test/java/org/openmrs/addonindex/legacy/LegacyControllerIT.java
```diff
@@ -68,7 +68,7 @@ public void setUp() throws Exception {
 		
 		AddOnInfoSummary summary = new AddOnInfoSummary(full);
 		
-		given(elasticSearchIndex.search(AddOnType.OMOD, "appui"))
+		given(elasticSearchIndex.search(AddOnType.OMOD, "appui",null))
```
djazayeri: always a space after a comma

src/test/java/org/openmrs/addonindex/legacy/LegacyControllerIT.java
```diff
@@ -88,7 +88,7 @@ public void setUp() throws Exception {
 	
 	@Test
 	public void testFindModulesWithoutVersion() throws Exception {
-		LegacyFindModulesResponse response = controller.findModules("appui", 0, 0, 123, null, null);
+		LegacyFindModulesResponse response = controller.findModules("appui",null, 0, 0, 123, null, null);
```
djazayeri: space after comma. And another one below.

src/test/java/org/openmrs/addonindex/rest/AddOnControllerIT.java
```diff
@@ -54,7 +54,7 @@ public void setUp() throws Exception {
 		info.setDescription("For reporting");
 		info.addVersion(version);
 		
-		when(index.search(null, "report")).thenReturn(singletonList(new AddOnInfoSummary(info)));
+		when(index.search(null, "report",null)).thenReturn(singletonList(new AddOnInfoSummary(info)));
```
djazayeri: space after comma

src/test/java/org/openmrs/addonindex/service/ElasticSearchIndexManualTest.java
```diff
@@ -61,7 +61,7 @@ public void testIndex() throws Exception {
 		Collection<AddOnInfoAndVersions> allByType = elasticSearchIndex.getAllByType(AddOnType.OMOD);
 		assertThat(allByType.size(), greaterThanOrEqualTo(1));
 		
-		Collection<AddOnInfoSummary> results = elasticSearchIndex.search(AddOnType.OMOD, "testing");
+		Collection<AddOnInfoSummary> results = elasticSearchIndex.search(AddOnType.OMOD, "testing",null);
```
djazayeri: space

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: I didn't really get what you meant here. But if you meant that this method is becoming confusing for a user who calls this function due to the ways of getting results by tag(i.e either itself or as a combination) then that's the reason I had used the getallbyTag method. Another way of removing the confusion would be to add a comment which clearly mentions what's happening. I should've done that already.

src/main/java/org/openmrs/addonindex/legacy/LegacyController.java
```diff
@@ -75,13 +75,14 @@ public String checkUpdate(@PathVariable("moduleId") String moduleId) throws Exce
 	@RequestMapping(method = RequestMethod.GET, value = "findModules")
 	public LegacyFindModulesResponse findModules(
 			@RequestParam(value = "sSearch", required = false) String query,
+			@RequestParam(value = "tags", required = false) String tag,
```
reubenvarghese1: Makes sense. I thought I'd do that but then I was like maybe we shall support tags in the future in OpenMRS especially since there's a Gsoc project already based on it. Moreover, I thought that since the parameter is optional, it wouldn't matter. But since you've mentioned, I'll change it's value to null and pass it

reubenvarghese1: Replying to your above comment, I actually saw that you didn't explicitly test the search function Inside the Elastic search test file. Moreover you tested the fact as to whether the search feature was working and hence I didn't ​write any tests either

src/test/java/org/openmrs/addonindex/service/ElasticSearchIndexManualTest.java
reubenvarghese1: You have sharp eyes 😛. I'll fix it :)

src/main/java/org/openmrs/addonindex/legacy/LegacyController.java
djazayeri: If there's someplace that's testing the search of `type` and `query` in any way, then you should add an equivalent search for tag. Otherwise, ignore it. (This will be the only time I tell you it's okay to ignore writing a test. But there's no worthwhile way of testing this.)

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
djazayeri: I mean that the caller might want to combine other parameters along with tag, like:

    index.search(Type.MODULE, null, "form-entry")
    index.search(Type.OWA, "html", "form-entry")

Your current code doesn't support this, but it would be easy to add.

reubenvarghese1: Ah okay, that's not an issue. I'll add it :)

src/main/ui/app/route/SearchPage.jsx
reubenvarghese1: This does have the same formatting. Actually it's showing it incorrectly here for some reason i.e. if you click on the above view, the formatting is fine... Strange but true :)

src/main/ui/app/route/SearchPage.jsx
reubenvarghese1: Yes they were

src/main/ui/app/route/SearchPage.jsx
reubenvarghese1: Woah...I honestly have no idea as to where that came from!

src/main/ui/app/route/Show.jsx
reubenvarghese1: I'll add to our list of tasks. This is a minor one and hopefully can be focused on and discussed upon later :)

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: Hopefully now the method can be called for all combinations

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -33,10 +33,8 @@ export default class SearchPage extends Component {
     doSearch() {
         const query = this.props.location.query.q;
         const type = this.props.location.query.type;
-        const searchKey = `type:${type ? type : "all"} query:${query ? query : ""}`;
-
-        console.log(searchKey);
-        console.log(this.state.latestSearch);
+	const tag = this.props.location.query.tag;
```
reubenvarghese1: Pls Note: This looks out of order for some reason but on viewing the file, it is perfectly indented

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -51,6 +49,9 @@ export default class SearchPage extends Component {
         if (query) {
             url += "&q=" + query;
         }
+	if (tag) {
```
reubenvarghese1: Again, this also looks out of order by one space but it is just fine if you view the file itself

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
djazayeri: I think it's still overcomplicated.

The pseudocode should look like

    if (type != null) {
        // add a clause to boolQB
    }
    if (tag != null) {
        // add a clause to boolQB
    }
    if (query != null) {
        // add lots of clauses to boolQB
    }
    // execute query and return results
You don't need to special-case `if (tag != null && query == null && type == null) {`

src/main/ui/app/route/SearchPage.jsx
djazayeri: I think it's because your IDE is using tabs instead of spaces.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: I had tried that first but for some reason I got many search results returned . For example: If I tried to get the modules with the tag "form-entry" , I got many results while actually I should get only 4 . Hence I added these clauses.

src/main/ui/app/route/SearchPage.jsx
reubenvarghese1: Nope, tried fixing that. Still not working

PR #18, opened 2017-06-25
AO 4 Fetch OpenMRS Core versions Rest API

reubenvarghese1:


src/main/java/org/openmrs/addonindex/domain/artifactory/ArtifactoryFolderInfoChild.java
```diff
@@ -0,0 +1,49 @@
...
+        return version;
+    }
+
+    public void setVersion(Version version) {
+        this.version = version;
+    }
+
+    @Override
+    public int compareTo(ArtifactoryFolderInfoChild o) {
+        return 0;
```
djazayeri: This is wrong. You should only make something implement Comparable if you are actually going to implement its compareTo method. In this case I don't see why these would be comparable since they don't have a natural ordering.

So, get rid of the compareTo method and remove the "implements Comparable"

src/main/java/org/openmrs/addonindex/domain/artifactory/ArtifactoryFolderInfoChild.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.addonindex.domain.artifactory;
+
+import org.openmrs.addonindex.util.Version;
+
+public class ArtifactoryFolderInfoChild implements Comparable<ArtifactoryFolderInfoChild> {
```
djazayeri: Add one sentence of javadoc indicating what this is for.

src/main/java/org/openmrs/addonindex/domain/artifactory/CoreVersionsSummary.java
```diff
@@ -0,0 +1,102 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.addonindex.domain.artifactory;
+
+import java.util.List;
+
+
+public class CoreVersionsSummary {
```
djazayeri: I think this class should be called ArtifactoryFolderInfo. And it should have a line of javadoc saying what it's for.

Further, I would prefer to remove any fields that we're not using. (It's also confusing to see created/lastModified/lastUpdated as Strings instead of some kind of date type.)

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Collection;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class VersionList {
+
+    private SortedSet<Version> versionSortedSet;
+
+    public VersionList(Collection<String> coreversions) {
+        String[] items = {"alpha", "beta", "hibernate", "maven", "RC", "SNAPSHOT"};
```
djazayeri: make this a "private static final" field at the class level, instead of instantiating it every time as a variable in the constructor.
Also, instead of "items" call it "STRINGS_TO_EXCLUDE".

djazayeri: Why are "hibernate" and "maven" here?

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,46 @@
...
+        this.versionSortedSet = new TreeSet<>();
+        for (String version : coreversions) {
+            ArtifactoryFolderInfoChild artifactoryFolderInfoChild = new ArtifactoryFolderInfoChild();
```
djazayeri: Why are you using ArtifactoryFolderInfoChild here? I would expect that the only purpose of that class is to support automatic deserialization of the JSON response we get from artifactory.

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.TreeSet;
+
+public class VersionList {
+
+    private SortedSet<Version> versionSortedSet;
+
+    public VersionList(Collection<String> coreversions) {
+        String[] items = {"alpha", "beta", "hibernate", "maven", "RC", "SNAPSHOT"};
+        this.versionSortedSet = new TreeSet<>();
+        for (String version : coreversions) {
```
djazayeri: FYI I would write the code like this, since I would consider it easier to read this way:

    for (String candidateVersion : coreversions) {
        if (!stringContainsItemFromList(candidateVersion, STRINGS_TO_EXCLUDE) {
            versionSortedSet.add(new Version(candidateVersion));
        }
    }

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,46 @@
...
+import org.openmrs.addonindex.util.Version;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class VersionList {
+
+    private SortedSet<Version> versionSortedSet;
```
djazayeri: Just call this "versions". If this naming were purely private within a method and you wanted to very explicitly say "this is sorted" then you could name it like this, but for someone who is consuming this class they shouldn't care at first glance that this is specifically a SortedSet (but if they do care they can just check the return type of getVersions).

src/main/java/org/openmrs/addonindex/rest/artifactory/CoreVersionsController.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.SortedSet;
+
+@RestController
+public class CoreVersionsController {
+
+    private VersionsService versionsService;
+
+    @Autowired
+    public CoreVersionsController(VersionsService versionsService){ this.versionsService = versionsService; }
```
djazayeri: formatting. :-)

    public CoreVersionsController(VersionsService versionsService) {
        this.versionsService = versionsService;
    }

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -0,0 +1,80 @@
...
+        String json;
+        if (strategy == Strategy.LOCAL) {
+            logger.debug("LOCAL strategy");
+            json = StreamUtils.copyToString(getClass().getClassLoader().getResourceAsStream("openmrs-core-versions.json"),
+                    Charset.defaultCharset());
+        } else {
+            logger.debug("FETCH strategy: " + url);
+            json = restTemplateBuilder.build().getForObject(url, String.class);
+        }
+        CoreVersionsSummary children;
```
djazayeri: "children" is not the right variable name.

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            children = mapper.readValue(json, CoreVersionsSummary.class);
+        } catch (Exception ex) {
+            throw new RuntimeException("File downloaded from " + url + " could not be parsed", ex);
```
djazayeri: Never throw just a RuntimeException. If you're doing this then you should throw either IllegalArgumentException or IllegalStateException. (the first one means that the method caller passed an illegal argument. the second means that some state that's external to this method is wrong.)

djazayeri: PS- a random google hit about this: https://www.securecoding.cert.org/confluence/display/java/ERR07-J.+Do+not+throw+RuntimeException%2C+Exception%2C+or+Throwable

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        if (logger.isInfoEnabled()) {
+            logger.info("We have " + children.getChildren().size() + " Core versions to index");
+        }
+        if (children.size() > 0) {
+            versionsService.setVersions(children.getChildren().stream().map(ArtifactoryFolderInfoChild::getUri).collect(Collectors.toList()));
```
djazayeri: These stream -> collection operations are quite hard to read and interpret for someone else. I recommend that you always put newlines and indentations such that they look like the Single-Line Statement that's in the _question_ at https://stackoverflow.com/questions/24649971/java-8-formatting-lambda-with-newlines-and-indentation

This makes reading significantly easier.

src/main/resources/application.yml
```diff
@@ -1,22 +1,31 @@
...
+  api_key:  # override in your local config
 
 scheduler:
+  fetch_core_versions_list:     # for each add-on we want to index, fetch its details
+    initial_delay: 5000      # wait 5 seconds to start
+    period: 14400000          # execute every 4 hours
+
   fetch_add_on_list:     # fetches our master list of add-ons that we should index (from github)
-    initial_delay: 5000  # wait 5 seconds to start
+    initial_delay: 9000  # wait 5 seconds to start
```
djazayeri: not sure why you're changing this but if you do then change the comment also

src/main/resources/application.yml
```diff
@@ -1,22 +1,31 @@
...
     period: 14400000     # execute every 4 hours
 
   fetch_details_to_index:     # for each add-on we want to index, fetch its details
-    initial_delay: 15000      # wait 15 seconds to start
+    initial_delay: 20000      # wait 15 seconds to start
```
djazayeri: not sure why you're changing this but if you do then change the comment also

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
reubenvarghese1: Because certain packages do have the term "hibernate" in them( Actually its not needed here as the version name also contains the term Snapshot but maybe in the future there might be a version with just hibernate in it's name. Let me know if it's not needed ) . There is also a file called maven.xml do not want that either .You may check https://openmrs.jfrog.io/openmrs/public/org/openmrs/api/openmrs-api/

src/main/resources/application.yml
reubenvarghese1: I'll revert that. Had changed it initially :)

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
reubenvarghese1: Awesome ... I just realised that I do so many things that are just not necessary!

reubenvarghese1:
@djazayeri Do you want me to change the exception in FetchAddOnlIst to IllegalStateExceprtion as well instead of runtime exception?

src/main/java/org/openmrs/addonindex/domain/artifactory/ArtifactoryFolderInfo.java
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.addonindex.domain.artifactory;
+
+import java.util.List;
+
+/**
+ * Represents the list of OpenMRS core versions we want to index.
```
djazayeri: I would say:
* represents the response from (link to the REST call)
* we use this to determine the list of OpenMRS core versions

src/main/java/org/openmrs/addonindex/domain/artifactory/ArtifactoryFolderInfoChild.java
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.addonindex.domain.artifactory;
+
+import org.openmrs.addonindex.util.Version;
+
+/**
+ * Represents a single version(uri) and folder status( whether the entry is a folder or not)
```
djazayeri: In retrospect, putting javadoc here doesn't really help. Just say "@see ArtifactoryFolderInfo"

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class VersionList {
+
+    private SortedSet<Version> versions;
+
+    private static final String[] STRINGS_TO_EXCLUDE = {"alpha", "beta", "hibernate", "maven", "RC", "SNAPSHOT"};
```
djazayeri: I would get rid of hibernate, unless we really need it.
To be more clear, you can exclude "maven.xml" (instead of just "maven"). But actually, couldn't we exclude things before they even get to this class if we include only where ArtifactoryFolderInfoChild.folder == true?

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public VersionList(Collection<String> coreversions) {
+        this.versions = new TreeSet<>();
+        for (String candidateVersion : coreversions) {
+            if (!stringContainsItemFromList(candidateVersion, STRINGS_TO_EXCLUDE)) {
+                this.versions.add(new Version(candidateVersion.replaceAll("/", "")));
```
djazayeri: Why would we replace / with an empty string?

First off, I would expect that we clean up the string in whatever method calls this constructor.
Secondly, if we're expecting this string to have a path beforehand, why not take a substring?

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -0,0 +1,85 @@
...
+            json = restTemplateBuilder.build().getForObject(url, String.class);
+        }
+        ArtifactoryFolderInfo versionlist;
+        try {
+            versionlist = mapper.readValue(json, ArtifactoryFolderInfo.class);
+        } catch (Exception ex) {
+            throw new IllegalStateException("File downloaded from " + url + " could not be parsed", ex);
+        }
+        if (logger.isInfoEnabled()) {
+            logger.info("We have " + versionlist.getChildren().size() + " Core versions to index");
```
djazayeri: Just say "There are xyz versions of openmrs-core"

src/main/java/org/openmrs/addonindex/domain/artifactory/VersionList.java
reubenvarghese1: The only method that calls this constructor is the VersionsService which was the one initially responsible for cleaning the string. If I am not mistaken , you had asked me to move it to versionlist , the reason being that every time the rest api would call a function, the string would be screened all over again which is not needed. Hence it's now modified such that it is only screened when the Scheduler class sets the versions

reubenvarghese1: Could you rephrase your second point please?

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        if (versionlist.size() > 0) {
+            versionsService.setVersions(versionlist
+                    .getChildren()
+                    .stream()
+                    .map(ArtifactoryFolderInfoChild::getUri)
+                    .collect(Collectors.toList())
```
djazayeri: This is where you should add a function that cleans up the uri and turns it into a core version name. (E.g. in my other comment I said "I don't see why you're doing this here.")

PR #19, opened 2017-06-28
AO 4 Fetch OpenMRS Core versions Rest API (Part 2) + UI

reubenvarghese1:
This is the final additions to the Rest API for AO 4

### DIFF
--- a/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
+++ b/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
@@ -203,7 +203,7 @@ public class AddOnInfoAndVersions {
             }
             return null;
         }
-        return getVersion(getLatestVersion()).orElseThrow(IllegalArgumentException::new);
+        return getVersion(getLatestVersion()).isPresent() ? getVersion(getLatestVersion()).get() : null;
     }
 	
 	public void setDetailsBasedOnLatestVersion() {
--- a/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
+++ b/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
 
+import org.openmrs.addonindex.util.OpenmrsVersionCompareUtil;
 import org.openmrs.addonindex.util.Version;
 
 import io.searchbox.annotations.JestId;
@@ -172,4 +173,16 @@ public class AddOnInfoAndVersions {
 		}
 		tags.add(tag);
 	}
+
+    public AddOnVersion getLatestSupportedVersion(String userCoreVersion) throws Exception {
+        if (userCoreVersion != null) {
+            for (AddOnVersion addOnVersion : getVersions()) {
+                if (OpenmrsVersionCompareUtil.matchRequiredVersions(userCoreVersion, addOnVersion.getRequireOpenmrsVersion())) {
+                    return addOnVersion;
+                }
+            }
+            return null;
+        }
+        return getVersion(getLatestVersion()).orElseThrow(IllegalArgumentException::new);
+    }
 }
--- a/src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
+++ b/src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
@@ -67,7 +67,7 @@ public class FetchCoreVersionsList {
             throw new IllegalStateException("File downloaded from " + url + " could not be parsed", ex);
         }
         if (logger.isInfoEnabled()) {
-            logger.info("There are " + versionlist.getChildren().size() + " versions to index");
+            logger.info("There are " + versionlist.getChildren().size() + " openmrs-core versions");
         }
         if (versionlist.size() > 0) {
             List<String> versions = new ArrayList<>();
@@ -83,7 +83,7 @@ public class FetchCoreVersionsList {
         }
     }
 
-    private static boolean stringContainsItemFromList(String inputStr, String[] items) {
+    public boolean stringContainsItemFromList(String inputStr, String[] items) {
         return Arrays.stream(items).parallel().anyMatch(inputStr::contains);
     }
 
--- a/src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
+++ b/src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
@@ -83,7 +83,7 @@ public class FetchCoreVersionsList {
         }
     }
 
-    public boolean stringContainsItemFromList(String inputStr, String[] items) {
+    private static boolean stringContainsItemFromList(String inputStr, String[] items) {
         return Arrays.stream(items).parallel().anyMatch(inputStr::contains);
     }
 
--- a/src/main/ui/app/route/Show.jsx
+++ b/src/main/ui/app/route/Show.jsx
@@ -132,11 +132,11 @@ export default class Show extends Component {
             if (this.state.latestVersion){
                 if (this.state.latestVersion.version === addon.versions[0].version){
                     version = <span>Download<br/>
-                        <small>Latest Version {this.state.latestVersion.version}</small></span>;
+                        <small>Latest Version: {this.state.latestVersion.version}</small></span>;
                 }
                 else {
                     version = <span>Download<br/>
-                        <small>Latest SupportedVersion {this.state.latestVersion.version}</small></span>;
+                        <small>Supported Version: {this.state.latestVersion.version}</small></span>;
                 }
             }
             else {
--- a/src/main/ui/app/route/Show.jsx
+++ b/src/main/ui/app/route/Show.jsx
@@ -31,7 +31,43 @@ export default class Show extends Component {
                 })
                 .catch(err => {
                     this.setState({error: err});
-                })
+                });
+
+        this.getLatestSupportedVersion(this.props.selectedVersion)
+    }
+
+    getLatestSupportedVersion(openmrsCoreVersion){
+        let url = '/api/v1/addon/' + this.props.params.uid + '/latestVersion';
+        if (openmrsCoreVersion) {
+            url += '?&coreversion=' + openmrsCoreVersion;
+        }
+        console.log("fetchign for version"+ openmrsCoreVersion);
+        fetch(url)
+            .then(response => {
+                if (response.status >= 400) {
+                    throw new Error(response.statusText);
+                }
+                else if (response.status === 204){
+                    console.log(response.status.toString());
+                    this.setState({latestModule: null});
+                }
+                else {
+                    return response.json();
+                }
+            })
+            .then(version => {
+                this.setState({latestModule: version});
+            })
+            .catch(err => {
+                this.setState({error: err});
+            })
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if(nextProps.selectedVersion !== this.props.selectedVersion){
+            this.getLatestSupportedVersion(nextProps.selectedVersion)
+            console.log(this.state.latestModule);
+        }
     }
 
     formatRequiredModules(version) {
@@ -54,7 +90,6 @@ export default class Show extends Component {
     }
 
     formatTags(addon) {
-        
         let statusClass = "default";
         let status = addon.status;
         switch (addon.status) {
@@ -83,7 +118,6 @@ export default class Show extends Component {
         </div>
     }
 
-
     render() {
         if (this.state && this.state.error) {
             return <div>{this.state.error}</div>
@@ -94,6 +128,10 @@ export default class Show extends Component {
             const requirementmeaning = (
                     <Tooltip id="tooltip"><strong>Minimum version of the OpenMRS Platform required.</strong></Tooltip>
             );
+            let version = this.state.latestModule ?
+                "Latest Version(" + this.state.latestModule.version + ")" :
+                "Version not supported";
+            let versionDownloadUri = this.state.latestModule ? this.state.latestModule.downloadUri : null;
             let hostedSection = "";
             let hosted = "";
             if (addon.hostedUrl) {
@@ -123,6 +161,7 @@ export default class Show extends Component {
                             </Col>
                         </Row>
                         <div className="col-md-12 col-sm-12 col-xs-12 left-margin">
+                            <div className="col-md-10 col-sm-10 col-xs-10">
                             <Table condensed>
                                 <colgroup>
                                     <col className="col-md-2"/>
@@ -153,6 +192,19 @@ export default class Show extends Component {
                                 </tr>
                                 </tbody>
                             </Table>
+                            </div>
+                            <div className="col-md-2 col-sm-2 col-xs-2">
+                                <a href={versionDownloadUri}>
+                                    <Button className="primary" bsStyle="primary" bsSize="large"
+                                            disabled={versionDownloadUri === null}>
+                                        <span>Download</span>
+                                        <br/>
+                                        <span>
+                                            {version}
+                                        </span>
+                                    </Button>
+                                </a>
+                            </div>
                         </div>
                         <div>
                             <Table condensed hover>
--- a/src/main/ui/app/route/Show.jsx
+++ b/src/main/ui/app/route/Show.jsx
@@ -131,12 +131,12 @@ export default class Show extends Component {
             let version = "";
             if (this.state.latestVersion){
                 if (this.state.latestVersion.version === addon.versions[0].version){
-                    version = <span>Download Latest<br/>
-                        <small>Version {this.state.latestVersion.version}</small></span>;
+                    version = <span>Download<br/>
+                        <small>Latest Version {this.state.latestVersion.version}</small></span>;
                 }
                 else {
-                    version = <span>Download Latest Supported<br/>
-                        <small>Version {this.state.latestVersion.version}</small></span>;
+                    version = <span>Download<br/>
+                        <small>Latest SupportedVersion {this.state.latestVersion.version}</small></span>;
                 }
             }
             else {
@@ -208,7 +208,7 @@ export default class Show extends Component {
                             </div>
                             <div className="col-md-3 col-sm-3 col-xs-3">
                                 <a href={versionDownloadUri}>
-                                    <Button className="primary" bsStyle="primary" bsSize="large"
+                                    <Button className="primary" bsStyle={versionDownloadUri ? "primary" : "default"} bsSize="large"
                                             disabled={versionDownloadUri === null}>
                                         {version}
                                     </Button>
--- a/src/main/ui/app/route/Show.jsx
+++ b/src/main/ui/app/route/Show.jsx
@@ -31,6 +31,7 @@ export default class Show extends Component {
                 })
                 .catch(err => {
                     this.setState({error: err});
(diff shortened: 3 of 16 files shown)


reubenvarghese1:
The reason I added the function to the service is by looking at a function you had added in Indexing service

src/main/ui/app/route/Show.jsx
```diff
@@ -154,53 +191,63 @@ export default class Show extends Component {
...
-                                                    <Button bsSize="small"
-                                                            href={v.renameTo ? `/api/v1/addon/${addon.uid}/${v.version}/download` : v.downloadUri}>
-                                                        <i className="fa fa-download"></i>
-                                                        Download
-                                                    </Button>
-                                                </td>
-                                            </tr>
-                                    )
-                                })}
+                        <div className="col-md-2 col-sm-2 col-xs-2 pull-right">
```
djazayeri: Why does this need a pull-right?

src/main/ui/app/route/Show.jsx
```diff
@@ -154,53 +191,63 @@ export default class Show extends Component {
...
+                            <Button href={version.downloadUri} bsStyle="primary" bsSize="large">
+                                <span className="buttonstyleheading">Download
+                                </span><br/>
+                                <span className="buttonstylesubtext">
+                                Version : {version.version}
```
djazayeri: I would expect us to say either "Latest Version (version number)" or "Latest Supported Version (version number)"

src/main/ui/app/sass/addonindex.scss
```diff
@@ -167,3 +167,16 @@ a:hover, a:visited, a:link, a:active {
   border-top: 1px lightgray solid;
   padding-top: 10px;
 }
+
+.buttonstyleheading{
```
djazayeri: definitely need to clean this up.

src/main/ui/package.json
```diff
@@ -23,7 +23,8 @@
     "react": "^15.4.2",
     "react-bootstrap": "^0.30.7",
     "react-dom": "^15.4.2",
-    "react-router": "^3.0.2"
+    "react-router": "^3.0.2",
+    "react-select" : "*"
```
djazayeri: Did you do a regular `npm install --save-dev`` for this? I wonder why it's showing * instead of a minimum version number.

src/test/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsListIT.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+import static org.junit.Assert.assertThat;
+import static org.mockito.BDDMockito.given;
+
+@SpringBootTest
+@RunWith(SpringRunner.class)
+public class FetchCoreVersionsListIT {
```
djazayeri: This class is not actually testing what it's supposed to be testing.

What you want to do here is to call the scheduled task, and then make sure it did the right thing. (In this case that means verifying that it called VersionService.setVersions(...), or whatever that method is called.)

djazayeri:
When I do `npm install` with this code I get this output:

```
MacBook-Pro-3:ui djazayer$ npm install
openmrs-contrib-addonindex-ui@1.0.0 /Users/djazayer/code/openmrs/openmrs-contrib-addonindex/src/main/ui
├── UNMET PEER DEPENDENCY react@15.4.2
├── UNMET PEER DEPENDENCY react-dom@15.4.2
└─┬ react-select@0.9.1 
  └── react-input-autosize@0.6.13 

npm WARN react-select@0.9.1 requires a peer of react@^0.14.1 but none was installed.
npm WARN react-select@0.9.1 requires a peer of react-dom@^0.14.1 but none was installed.
npm WARN openmrs-contrib-addonindex-ui@1.0.0 No repository field.
```

djazayeri:
When I checked out the code and ran it, it does not work. I can see the widget, but it doesn't work right.

First, trying to make a selection fails inconsistently. If I make a selection by pressing Enter after selecting something, by typing +/- using the arrows, about half of the time it just goes back to showing Default. I can't figure out any pattern to it.

When I make a selection, and then I navigate to another route by clicking a link, then the widget loses its value and goes back to showing Default. (Though when I inspect the state with the ReactJS chrome plugin it does show that the state is preserved.

Most importantly, it doesn't seem to work at all. E.g. if I select version 1.6.4 and then I go to the idgen module, it still shows Download Latest Version 4.4.1, but it should only let me get version 3.2.

Also, UI-wise, the select box needs some kind of label to tell the user what it's for.

src/main/ui/app/route/Show.jsx
```diff
@@ -154,53 +190,66 @@ export default class Show extends Component {
...
-                                    )
-                                })}
+                        <div className="col-md-2 col-sm-2 col-xs-2">
+                            <a href={version.downloadUri}>
+                            <Button className="primary" bsStyle="primary" bsSize="large"
+                                    disabled={version.version === null}>
+                                <span>Download
+                                </span><br/>
+                                <span>
+                                  Latest Version : {version.version === null ? "Null" : version.version}
```
djazayeri: see my comment from previous review:

> I would expect us to say either "Latest Version (version number)" or "Latest Supported Version (version number)"

Also, never display "null" to an end user.

src/main/ui/package.json
```diff
@@ -23,7 +23,8 @@
     "react": "^15.4.2",
     "react-bootstrap": "^0.30.7",
     "react-dom": "^15.4.2",
-    "react-router": "^3.0.2"
+    "react-router": "^3.0.2",
+    "react-select": "*"
```
djazayeri: How did you install this? It should be showing a specific version. When I do `npm install --save react-select` it _does_ update this file to show `^0.9.1` instead of *

src/test/java/org/openmrs/addonindex/VersionServiceIT.java
```diff
@@ -0,0 +1,56 @@
...
+        a.setDescription("This is a test");
+        a.addVersion(v);
+
+        when(index.getByUid("testing-module")).thenReturn(a);
+
+        AddOnVersion addOnVersion = versionsService.getMatchedVersion("2.1.0", "testing-module");
+        assertThat(addOnVersion.getVersion().toString(), is("1.0"));
+
+        AddOnVersion nulladdOnVersion = versionsService.getMatchedVersion("1.6.3", "testing-module");
+        assertNull(null, nulladdOnVersion.getVersion());
```
djazayeri: I would expect that nulladdOnVersion == null.

src/test/java/org/openmrs/addonindex/VersionServiceIT.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.when;
+
+@SpringBootTest
+@RunWith(SpringRunner.class)
+public class VersionServiceIT
```
djazayeri: You have structured this IT correctly. Good work!

However it doesn't make sense to actually be integration-testing this as part of the service since all of the logic is carried out by the domain object. It's sufficient to unit test this in the add on class. (We can also have an end-to-end test in the controller, but we don't need this tested in all three places.)

src/test/java/org/openmrs/addonindex/domain/AddOnInfoAndVersionsTest.java
```diff
@@ -16,29 +17,43 @@
...
+        String json = mapper.writeValueAsString(info);
+        AddOnInfoAndVersions parsed = mapper.readValue(json, AddOnInfoAndVersions.class);
+        assertThat(parsed.getName(), is("OpenMRS"));
+        assertThat(parsed.getDescription(), is("Write code, save lives"));
+        assertThat(parsed.getVersions().size(), is(1));
+        assertThat(parsed.getVersions().get(0).getVersion().toString(), is("1.0"));
+    }
+
+    @Test
+    public void testGetLatestMatchedVersion() throws Exception {
```
djazayeri: In _this_ test class you should have 3 tests for different scenarios:

1. the latest version matches
2. some earlier version matches
3. no version matches

src/test/java/org/openmrs/addonindex/domain/AddOnInfoAndVersionsTest.java
```diff
@@ -16,29 +17,43 @@
...
+
+        AddOnVersion version = new AddOnVersion();
+        version.setVersion(new Version("1.0"));
+        version.setRequireOpenmrsVersion("2.1.0");
+
+        AddOnInfoAndVersions info = new AddOnInfoAndVersions();
+        info.setName("OpenMRS");
+        info.setDescription("Write code, save lives");
+        info.addVersion(version);
+        assertEquals("1.0", info.getLatestSupportedVersion("2.1.0").getVersion().toString());
```
djazayeri: The correct assertion here would be to check that what's returned here is the same object as you've defined above, not just comparing the string for equality.

    assertEquals(version, info.getLatestSupportedVersion("2.1.0"));

or the more modern way is

    assertThat(info.getLatestSupportedVersion("2.1.0"), is(version));

src/test/java/org/openmrs/addonindex/rest/CoreVersionsControllerIT.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        List<String> versions = new ArrayList<>();
+        versions.add("1.6.3");
+        versions.add("1.6.4");
+        VersionList versionList;
+        versionList = new VersionList(versions);
+        doReturn(versionList).when(versionsService).getVersions();
+
```
djazayeri: no extraneous blank lines!

src/test/java/org/openmrs/addonindex/rest/CoreVersionsControllerIT.java
```diff
@@ -0,0 +1,91 @@
...
+    }
```
djazayeri: but this needs a blank line after it!

src/test/java/org/openmrs/addonindex/rest/CoreVersionsControllerIT.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    public void getCoreversions() throws Exception {
+        ResponseEntity<String> entity = testRestTemplate.getForEntity("http://localhost:" + port + "/api/v1/coreversions",
+                String.class);
+
+        assertThat(entity.getStatusCode(), is(HttpStatus.OK));
+        assertThat(entity.getBody(),is("[\"1.6.3\",\"1.6.4\"]"));
```
djazayeri: You should use JSONAssert with strict=false, because we don't care about exact string equality. (see AddOnControllerIT)

src/test/java/org/openmrs/addonindex/rest/CoreVersionsControllerIT.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void getLatestSupportedVersionTest() throws Exception{
```
djazayeri: Don't need "Test" in the method name of the test.

djazayeri: From the perspective of testing strategy, this test is bad.

The value of having a (slow, expensive) integration test on a controller is to do an end-to-end test of the system. But in this case you're mocking VersionsService so we don't actually get an end-to-end test out of this. Instead we get a slow test that's retesting the same behavior that should be tested in the unit test for AddOnInfoAndVersions.

Instead of mocking the behavior of `versionsService.getMatchedVersion("2.1.0","testing-module")` you should have a @MockBean Index and mock the `index.getByUid("testing-module")`.

Also, a side-comment is that you're setting up _more_ test data in this test as compared to AddOnInfoAndVersionsTest. Instead you should be testing more of the business logic in that (faster, cheaper) unit test, and in this integration test you just test a single happy path.
(I'm not saying to remove one of the versions from this test, just that you should add more versions in the other test.)

src/test/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsListIT.java
```diff
@@ -0,0 +1,65 @@
...
+    @Autowired
+    private FetchCoreVersionsList fetchCoreVersionsList;
+
+    private VersionList versionList;
+
+    @Before
+    public void setUp() throws Exception {
+        when(restTemplateBuilder.build()).thenReturn(restTemplate);
+
+        ReflectionTestUtils.setField(fetchCoreVersionsList,
```
djazayeri: Why do you have to do this? Isn't this already the default behavior that's in application.yml?

src/test/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsListIT.java
```diff
@@ -0,0 +1,65 @@
...
+                "url",
+                "https://openmrs.jfrog.io/openmrs/api/storage/public/org/openmrs/api/openmrs-api/");
+        ReflectionTestUtils.setField(fetchCoreVersionsList,
+                "strategy",FetchCoreVersionsList.Strategy.FETCH);
+
+    }
+
+    @Test
+    public void testCoreVersionsJsonHandle() throws Exception{
```
djazayeri: The class that you're testing here is the one that has the business logic for excluding alpha/beta/SNAPSHOT/etc. Therefore your assertions in this test need to ensure that it's doing the right thing.

The correct thing to do would be to extract out a "filter" method, and to have another test class that's not an IT which has multiple @Test methods which test all the various behaviors.
The quick fix is to add more assertions to this test method to make sure you're excluding unwanted versions.

src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
```diff
@@ -24,152 +25,169 @@
...
+            if (userCoreVersion.equals("Default")){
+                return getVersion(getLatestVersion()).get();
+            }
+            else {
+                for (AddOnVersion addOnVersion : getVersions()) {
+                    if (OpenmrsVersionCompareUtil.matchRequiredVersions(userCoreVersion, addOnVersion.getRequireOpenmrsVersion())) {
+                        return addOnVersion;
+                    }
+                }
+                return new AddOnVersion();
```
djazayeri: I guess that in the controller you'd have to return a different HTTP Status code on null (example: https://stackoverflow.com/questions/32396884/return-http-204-on-null-with-spring-restcontroller)

And then in the react code that calls this controller you'd need to check the HTTP status before constructing any object.

.gitignore
```diff
@@ -4,6 +4,7 @@
 target
 config
 src/main/ui/node_modules
+src/main/ui/package-lock.json
```
djazayeri: If you look at the link I included it says:

> This file is intended to be committed into source repositories

Therefore we should include it!

src/test/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsListIT.java
djazayeri: When I comment out these lines the test passes for me.

Can you check if that's the case for you too? (If not it's not a problem to leave these lines in.)

src/main/ui/app/route/Show.jsx
```diff
@@ -94,6 +128,10 @@ export default class Show extends Component {
             const requirementmeaning = (
                     <Tooltip id="tooltip"><strong>Minimum version of the OpenMRS Platform required.</strong></Tooltip>
             );
+            let version = this.state.latestModule ?
```
djazayeri: I think the best UI here would be for this button to show one of 3 different things:

1. If you're going to get the very latest latest version, it should say: `Download Latest<br/><small>Version 1.2.0</small>`

2. If the latest version supported by your openmrs-core version is not the very latest, say: `Download Latest Supported<br/><small>Version 2.8.0</small>`

3. If no version is supported by your openmrs-core version: have a disabled button that says `No module version supports OpenMRS Core 1.2.0`

src/test/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsListIT.java
```diff
@@ -0,0 +1,68 @@
...
+                .thenReturn(getFileAsString("core-versions.json"));
+
+        fetchCoreVersionsList.fetchCoreVersionsList();
+        versionList = versionsService.getVersions();
+        assertNotNull(versionList);
+        assertThat(versionList.getVersions().first().toString(),is("1.6.3"));
+    }
+
+    @Test
+    public void testStringContainsItemFromList(){
```
djazayeri: Instead of (or in addition to) this you should test that the actual fetchCoreVersionList() method excludes the snapshots.
E.g. you can assert that versionList.getVersions() does not contain "1.9.12-SNAPSHOT"

src/main/java/org/openmrs/addonindex/scheduled/FetchCoreVersionsList.java
```diff
@@ -83,7 +83,7 @@ public void fetchCoreVersionsList() throws Exception {
         }
     }
 
-    private static boolean stringContainsItemFromList(String inputStr, String[] items) {
+    public boolean stringContainsItemFromList(String inputStr, String[] items) {
```
reubenvarghese1: So that I could test the function within the tests

djazayeri:
@reubenvarghese1 unfortunately this (still) does not work for me at all.

Whenever I try to select a platform version nothing happens.

djazayeri:
Further comments:

1. Reverse the order of the openmrs platform versions being displayed, so that the most recent is first.
2. Now that I see how my suggestions for the text to put on the buttons looks, I don't like it. For the normal case, go ahead and just have "Download" on the first line, and everything else on the smaller second line.
3. I think it's confusing that the only text field in the UI (when you're looking at a module) is this one. I.e. it makes me think this should be the search field. I would prefer if we had a link that says "Specify your OpenMRS Core/Platform version", which then turns into the autosuggest, and when you select one it goes back to looking like text (with a "change") link.
  - but it's not so important to do this now, let's get it working at all first!

reubenvarghese1:
@djazayeri  I don't really think that the select box is confusing. My reasons are : 
a) It's placeholder clearly mentions "Select platform version"
b) Suppose the user misses that. On placing his/her cursor inside the select box, automatically all the core versions show up and the user has to search for his platform version. Now suppose he starts typing a module name then, the user cannot choose anything(It just says no results found)

If you still feel that we need to add another link then we shall do so :)

src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
```diff
@@ -192,6 +193,18 @@ public void addTag(String tag) {
...
+    public AddOnVersion getLatestSupportedVersion(String userCoreVersion) throws Exception {
+        if (userCoreVersion != null) {
+            for (AddOnVersion addOnVersion : getVersions()) {
+                if (OpenmrsVersionCompareUtil.matchRequiredVersions(userCoreVersion, addOnVersion.getRequireOpenmrsVersion())) {
+                    return addOnVersion;
+                }
+            }
+            return null;
+        }
+        return getVersion(getLatestVersion()).orElseThrow(IllegalArgumentException::new);
```
djazayeri: Just return null in this case, don't throw an exception

PR #20, opened 2017-07-27
AO 6 - Add additional rest end points

reubenvarghese1:
List of changes:
1) Migrated all modules to use Bintray
2) Added two new fields called modulePackage and moduleId
3) Added new rest end points for accessing the modules using the above fields
4) Added tests for the same

### DIFF
--- a/src/main/java/org/openmrs/addonindex/rest/AddOnController.java
+++ b/src/main/java/org/openmrs/addonindex/rest/AddOnController.java
@@ -41,6 +41,12 @@ public class AddOnController {
 						   @RequestParam(value = "tag", required = false) String tag) throws Exception {
 		return index.search(type, query, tag);
 	}
+
+	@RequestMapping(method = RequestMethod.GET, value = "/api/v1/addon", params = "modulePackage")
+	public ResponseEntity<AddOnInfoAndVersions> getByModulePackage(@RequestParam(value = "modulePackage") String modulePackage) throws Exception {
+		AddOnInfoAndVersions addOn = index.getByModulePackage(modulePackage);
+		return new ResponseEntity<>(addOn, addOn == null ? HttpStatus.NOT_FOUND : HttpStatus.OK);
+	}
 	
 	@RequestMapping(method = RequestMethod.GET, value = "/api/v1/addon/{uid}")
 	public ResponseEntity<AddOnInfoAndVersions> getOne(@PathVariable String uid) throws Exception {
--- a/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
+++ b/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
@@ -123,9 +123,8 @@ public class FetchDetailsToIndex {
 	}
 	
 	void fetchExtraDetailsForEachVersion(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
-        boolean idAndPackageSet;
+        boolean alreadySetIdAndPackage = false;
 		AddOnInfoAndVersions existingInfo = indexingService.getByUid(toIndex.getUid());
-		idAndPackageSet = false;
 		for (ListIterator<AddOnVersion> iter = infoAndVersions.getVersions().listIterator(); iter.hasNext(); ) {
 			AddOnVersion version = iter.next();
 			if (existingInfo != null) {
@@ -148,8 +147,8 @@ public class FetchDetailsToIndex {
 					if (configXml == null) {
 						throw new IllegalArgumentException("No config.xml file in " + version.getDownloadUri());
 					} else {
-						handleConfigXml(configXml, version, infoAndVersions, idAndPackageSet);
-						idAndPackageSet = true;
+						handleConfigXml(configXml, version, infoAndVersions, alreadySetIdAndPackage);
+						alreadySetIdAndPackage = true;
 					}
 				}
 			}
@@ -178,7 +177,7 @@ public class FetchDetailsToIndex {
 	}
 	
 	void handleConfigXml(String configXml, AddOnVersion addOnVersion,
-                         AddOnInfoAndVersions addOnInfoAndVersions, Boolean idAndPackageSetStatus) throws Exception {
+                         AddOnInfoAndVersions addOnInfoAndVersions, boolean setModuleIdAndPackage) throws Exception {
 		// sometimes this says something like <!DOCTYPE ... "../lib-common/config-1.0.dtd">
 		// we don't need DTD validation in any case
 		configXml = configXml.replaceAll("<!DOCTYPE .*?>", "");
@@ -188,7 +187,8 @@ public class FetchDetailsToIndex {
 		handleRequireOpenmrsVersion(addOnVersion, xpath, config);
 		handleRequireModules(addOnVersion, xpath, config);
 		handleSupportedLanguages(addOnVersion, xpath, config);
-		if (!idAndPackageSetStatus){
+		if (!setModuleIdAndPackage){
+		    //The module versions are already sorted such that latest version is fetched first
 		    handleModuleIdAndPackage(addOnInfoAndVersions, xpath, config);
         }
 	}
@@ -230,7 +230,6 @@ public class FetchDetailsToIndex {
 
 	private void  handleModuleIdAndPackage(AddOnInfoAndVersions addOnInfoAndVersions, XPath xpath,
                                            Document config) throws XPathExpressionException{
-	    logger.info("Setting Info Hopefully Once");
         Object modulePackage = xpath.evaluate("/module/package/text()", config, XPathConstants.STRING);
         Object moduleId = xpath.evaluate("/module/id/text()", config, XPathConstants.STRING);
         if (StringUtils.hasText((String) modulePackage)) {
--- a/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
+++ b/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
@@ -63,7 +63,7 @@ public class FetchDetailsToIndex {
 	private RestTemplateBuilder restTemplateBuilder;
 	
 	private DocumentBuilderFactory documentBuilderFactory;
-	
+
 	@Autowired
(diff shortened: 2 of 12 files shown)


reubenvarghese1:
I have no idea as to why these little formatting changes are showing up :(

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
```diff
@@ -41,7 +41,13 @@ public AddOnController(Index index) {
 						   @RequestParam(value = "tag", required = false) String tag) throws Exception {
 		return index.search(type, query, tag);
 	}
-	
+
+	@RequestMapping(method = RequestMethod.GET, value = "/api/v1/addon", params = "modulePackage")
+	public ResponseEntity<AddOnInfoAndVersions> getByModulePackage(@RequestParam(value = "modulePackage", required = false) String modulePackage) throws Exception {
```
djazayeri: this should be `required=true`.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -156,7 +156,18 @@ public AddOnInfoAndVersions getByUid(String uid) throws IOException {
...
+
+	@Override
+	public AddOnInfoAndVersions getByModulePackage(String modulePackage) throws IOException {
+		return client.execute(new Search.Builder(new SearchSourceBuilder()
+				.size(1)
+				.query(QueryBuilders.matchQuery("modulePackage", modulePackage)).toString())
+				.addIndex(AddOnInfoAndVersions.ES_INDEX)
+				.build()).getHits(AddOnInfoAndVersions.class).stream()
+				.map(sr -> sr.source)
+				.collect(Collectors.toList()).get(0);
```
djazayeri: At first glance I expect this will throw an ArrayIndexOutOfBoundsException if there are no hits.
Can you verify via the ManualTest class, and fix this. (I.e. it should return null, not throw an exception, if you give an incorrect modulePackage.)

src/main/resources/add-ons-to-index.json
```diff
@@ -22,7 +24,9 @@
       }
     },
     {
-      "uid": "org.openmrs.module.accesslogging",
+      "uid": "org.omodopenmrs.module.accesslogging",
+      "moduleId": "1",
```
djazayeri: This is not right. The moduleId of this module is `accesslogging`. See for example https://github.com/openmrs/openmrs-module-htmlformentry/blob/3.3.2/omod/src/main/resources/config.xml#L5 . By convention this is always going to be the part of "modulePackage" that comes after the "org.openmrs.module.", and you can write a script to update this file reflecting that convention.

src/main/resources/add-ons-to-index.json
```diff
@@ -22,7 +24,9 @@
       }
     },
     {
-      "uid": "org.openmrs.module.accesslogging",
+      "uid": "org.omodopenmrs.module.accesslogging",
```
djazayeri: Why change the uid here?

src/main/resources/add-ons-to-index.json
```diff
@@ -34,27 +38,35 @@
...
+      "backend": "org.openmrs.addonindex.backend.Bintray",
+      "bintrayPackageDetails": {
+        "owner": "openmrs",
+        "repo": "omod",
+        "package": "accesslogging"
       }
     },
     {
       "uid": "org.openmrs.module.addgutteritem",
+      "moduleId": "2",
```
djazayeri: like my comment above, the moduleId should be `addgutteritem`. And same goes for all the rest.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
reubenvarghese1: I checked this ,you're correct. I'll fix it rightaway

djazayeri:
Two things:

1. Don't you still need to update /src/main/resources/elasticsearch/addOnInfoAndVersions-mappings.json to reflect the new fields, since they're supposed to be treated as exact matches?

2. Sorry for not noticing this earlier, but it's actually a mistake to put all of the moduleId and modulePackage in the addons-to-index file, and these fields don't belong on AddOnToIndex either. This is info that can be fetched from the config.xml file in the modules when we download and inspect them, so we should avoid repeating it in the json file. Specifically, in org.openmrs.addonindex.scheduled.FetchDetailsToIndex#fetchExtraDetailsForEachVersion you should add another method all next to handleConfigXml, or change that method itself, so that we can set these properties on AddOnInfoAndVersions based on getting them from config.xml.

dkayiwa:
@reubenvarghese1 how is this going? Looks like we need it for the manage modules OWA. 😊

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
```diff
@@ -147,7 +148,8 @@ void fetchExtraDetailsForEachVersion(AddOnToIndex toIndex, AddOnInfoAndVersions
 					if (configXml == null) {
 						throw new IllegalArgumentException("No config.xml file in " + version.getDownloadUri());
 					} else {
-						handleConfigXml(configXml, version);
+						handleConfigXml(configXml, version, infoAndVersions, idAndPackageSet);
+						idAndPackageSet = true;
```
djazayeri: Ideally we want to make sure that we do this only for the most recent version.
If this iteration is already sorted, then what you've done here is fine, but just add a comment saying that this is already sorted with the latest version first so we just do this the first time.

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
```diff
@@ -123,8 +123,9 @@ void getDetailsAndIndex(AddOnToIndex toIndex) throws Exception {
 	}
 	
 	void fetchExtraDetailsForEachVersion(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
+        boolean idAndPackageSet;
```
djazayeri: Call this variable `alreadySetIdAndPackage`

And also, you can assign it in the same line, e.g. `boolean alreadySetIdAndPackage = false`

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
reubenvarghese1: I added the boolean just because the iteration is already sorted and the latest version is fetched first

reubenvarghese1:
@dkayiwa We are currently working on this PR actively. The feature has more or less been implemented and the only thing left is a review. I will let you know as soon as this PR is merged. 
Thanks for your patience :-)

src/test/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndexTest.java
```diff
@@ -49,20 +53,32 @@ public void testParsingConfigXmlForRequiredModuleVersion() throws Exception {
...
+
+    @Test
+	public void testParsingConfigXmlForSettingModulePackageAndId() throws Exception {
+	    FetchDetailsToIndex task = new FetchDetailsToIndex(null, null);
+	    AddOnVersion version = new AddOnVersion();
+	    AddOnInfoAndVersions addOnInfoAndVersions = new AddOnInfoAndVersions();
+	    task.handleConfigXml(getFileAsString("config.withRelativePathDtd.xml"), version, addOnInfoAndVersions, false);
+	    assertThat(addOnInfoAndVersions.getModulePackage(), is("org.openmrs.module.mdrtb"));
+	    assertThat(addOnInfoAndVersions.getModuleId(), is("mdrtb"));
+	}
```
djazayeri: This test is good.

So you know, for completeness you would also want to have a test where:

* there are two versions
* the moduleId or modulePackage changes between them
* the test ensures that the id/package are taken from the latest version, not the earliest version

The philosophy of test-driven development, or behavior-driven development, says that if this behavior is in the code, you should have started by writing a test for it. :-)

(I'm not going to hold up merging the PR for this, but you're welcome to add it in another commit.)

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
```diff
@@ -175,7 +177,8 @@ String fetchConfigXml(AddOnToIndex addOnToIndex, AddOnVersion addOnVersion) thro
 		return null;
 	}
 	
-	void handleConfigXml(String configXml, AddOnVersion addOnVersion) throws Exception {
+	void handleConfigXml(String configXml, AddOnVersion addOnVersion,
+                         AddOnInfoAndVersions addOnInfoAndVersions, boolean setModuleIdAndPackage) throws Exception {
```
djazayeri: Sorry, my previous comment was unclear/misleading. I had meant that you should have this new argument represent "should this method call also set the id and package". However since you're using it the other way (i.e. true means it's already done) then you should change this variable name to be "skipModuleIdAndPackage".

djazayeri:
@reubenv, I am going to merge this now. Please do one more followup commit to fix the variable name in the handleConfigXml method. You may optionally also add another test for the change-module-id-between-versions scenario.

But you can go ahead and commit this followup fix directly, you don't need to wait for me to look at a PR.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -156,6 +158,24 @@ public AddOnInfoAndVersions getByUid(String uid) throws IOException {
 		return client.execute(new Get.Builder(AddOnInfoAndVersions.ES_INDEX, uid).build())
 				.getSourceAsObject(AddOnInfoAndVersions.class);
 	}
+
+	@Override
+	public AddOnInfoAndVersions getByModulePackage(String modulePackage) throws IOException {
+		SearchResult result = client.execute(new Search.Builder(new SearchSourceBuilder()
+				.size(1)
+				.query(QueryBuilders.matchQuery("modulePackage", modulePackage)).toString())
```
djazayeri: I think there should be a clause here to limit this to only type=OMOD. Please add this as a one-liner in a followup commit. (You don't need my review on that.)

PR #21, opened 2017-08-17
AO 6 Fixed rest end points issue

reubenvarghese1:
Fixed rest end points issue

### DIFF
--- a/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
+++ b/src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
@@ -192,4 +192,9 @@ public class AddOnInfoAndVersions {
 		}
 		tags.add(tag);
 	}
+
+	public void setDetailsBasedOnLatestVersion(AddOnVersion version){
+	    setModuleId(version.getModuleId());
+	    setModulePackage(version.getModulePackage());
+    }
 }
--- a/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
+++ b/src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
@@ -111,7 +111,9 @@ public class FetchDetailsToIndex {
 			if (fetchExtraDetails) {
 				fetchExtraDetailsForEachVersion(toIndex, infoAndVersions);
 			}
-			
+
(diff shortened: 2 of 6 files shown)


src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
```diff
@@ -192,4 +192,9 @@ public void addTag(String tag) {
 		}
 		tags.add(tag);
 	}
+
+	public void setDetailsBasedOnLatestVersion(AddOnVersion version){
```
djazayeri: Move the logic for picking the latest version of this module into this method. I.e. it should take no arguments, and it should get the latest version itself.

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
```diff
@@ -111,7 +111,9 @@ void getDetailsAndIndex(AddOnToIndex toIndex) throws Exception {
 			if (fetchExtraDetails) {
 				fetchExtraDetailsForEachVersion(toIndex, infoAndVersions);
 			}
-			
+
+			infoAndVersions.setDetailsBasedOnLatestVersion(infoAndVersions.getVersion(
```
djazayeri: Move the logic for choosing the latest version into AddOnInfoAndVersion. Actually, it seems worth adding getLatestAddOnVersion() would make sense, since I feel like we've tried to do this a couple of times now.

src/test/java/org/openmrs/addonindex/domain/AddOnInfoAndVersionsTest.java
```diff
@@ -40,5 +39,23 @@ public void testSerialization() throws Exception {
...
+    public void setDetailsBasedOnLatestVersion(){
+        AddOnVersion version = new AddOnVersion();
+        version.setVersion(new Version("2.0"));
+        version.setModuleId("1");
+        version.setModulePackage("org.openmrs.module.openmrs");
+
+        AddOnInfoAndVersions info = new AddOnInfoAndVersions();
+        info.setName("OpenMRS");
+        info.setDescription("Write code, save lives");
+        info.addVersion(version);
```
djazayeri: After making the changes I mentioned earlier in the code review, you'll want to have this test put _two_ versions on AddOnInfoAndVersions and ensure it uses the right one.

PR #22, opened 2017-08-19
AO 6 add new rest end points latest

reubenvarghese1:


src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
```diff
@@ -192,4 +192,10 @@ public void addTag(String tag) {
 		}
 		tags.add(tag);
 	}
+
+	public void setDetailsBasedOnLatestVersion(){
+        AddOnVersion version = getVersion(getLatestVersion()).orElseThrow(IllegalArgumentException::new);
```
djazayeri: If we're going to throw an exception it should be IllegalStateException. But actually, let's just return in this case. There's no need to fail here.

src/test/java/org/openmrs/addonindex/domain/AddOnInfoAndVersionsTest.java
```diff
@@ -40,5 +39,29 @@ public void testSerialization() throws Exception {
 		assertThat(parsed.getVersions().size(), is(1));
 		assertThat(parsed.getVersions().get(0).getVersion().toString(), is("1.0"));
 	}
+
+	@Test
+    public void setDetailsBasedOnLatestVersion(){
```
djazayeri: nice test!

djazayeri:
These commits speak about a ticket AO-6, but that ticket isn't actually the relevant one. I couldn't find the right ticket in fact, so I just created AO-9 for this.

PR #29, opened 2017-12-14
AO 14 Add download counts support for AddOns

reubenvarghese1:
This commit aims at adding download counts support to AddOns.

### DIFF
--- a/src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
+++ b/src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
@@ -30,6 +30,5 @@ public interface BackendHandler {
 	 * @throws Exception
 	 */
 	AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) throws Exception;
-	void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception;
 	
 }
--- a/src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
+++ b/src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
@@ -30,5 +30,6 @@ public interface BackendHandler {
 	 * @throws Exception
 	 */
 	AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) throws Exception;
+	void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception;
 	
 }
--- a/src/main/java/org/openmrs/addonindex/backend/Bintray.java
+++ b/src/main/java/org/openmrs/addonindex/backend/Bintray.java
@@ -71,7 +71,7 @@ public class Bintray implements BackendHandler, SupportsDownloadCounts {
 	}
 
 	@Override
-	public Integer fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
+	public void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
 		//We won't be checking for existing info as download counts is dynamic
 		// and has to be fetched each time
 		logger.info("Fetching Download Counts for " + toIndex.getUid());
@@ -88,7 +88,6 @@ public class Bintray implements BackendHandler, SupportsDownloadCounts {
 		catch (Exception ex) {
 			logger.error("Error fetching details for" + toIndex.getUid(), ex);
 		}
-		return totalDownloadCounts;
 	}
 
 	AddOnInfoAndVersions handlePackageJson(AddOnToIndex addOnToIndex, String packageJson) throws IOException {
--- a/src/main/java/org/openmrs/addonindex/backend/Bintray.java
+++ b/src/main/java/org/openmrs/addonindex/backend/Bintray.java
@@ -33,7 +33,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 @Component
-public class Bintray implements BackendHandler {
+public class Bintray implements BackendHandler, SupportsDownloadCounts {
 	
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 	
@@ -71,32 +71,24 @@ public class Bintray implements BackendHandler {
 	}
 
 	@Override
-	public void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
+	public Integer fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
 		//We won't be checking for existing info as download counts is dynamic
 		// and has to be fetched each time
 		logger.info("Fetching Download Counts for " + toIndex.getUid());
 		BintrayDownloadCounts downloadCounts;
+		Integer totalDownloadCounts = 0;
 		String json;
-		BintrayPackageDetails bintrayPackageDetails = toIndex.getBintrayPackageDetails();
-		String baseurl = "https://bintray.com/statistics/packageGeoStats?&pkgPath=/" + bintrayPackageDetails.getOwner()
-				+ "/" + bintrayPackageDetails.getRepo() + "/" + bintrayPackageDetails.getPackageName();
+		String downloadStatsUrl = downloadCountUrlFor(toIndex);
 		try {
-			json = restTemplateBuilder.build().getForObject(baseurl, String.class);
+			json = restTemplateBuilder.build().getForObject(downloadStatsUrl, String.class);
 			downloadCounts = mapper.readValue(json, BintrayDownloadCounts.class);
-			Integer totalDownloadCounts = 0;
-			for (Integer downloadCount : downloadCounts.getTotalDownloads().values()) {
-				try {
-					totalDownloadCounts += downloadCount;
-				}
-				catch (Exception e) {
-					logger.error("Error getting details for " + toIndex.getUid(), e);
-				}
-			}
-			infoAndVersions.setDownloadCounts(totalDownloadCounts);
+			totalDownloadCounts = handleBintrayGeoStats(toIndex, downloadCounts);
+			infoAndVersions.setDownloadCountInLast30Days(totalDownloadCounts);
 		}
 		catch (Exception ex) {
-			throw new RuntimeException("Download counts from " + baseurl + " could not be parsed", ex);
+			logger.error("Error fetching details for" + toIndex.getUid(), ex);
 		}
+		return totalDownloadCounts;
 	}
 
 	AddOnInfoAndVersions handlePackageJson(AddOnToIndex addOnToIndex, String packageJson) throws IOException {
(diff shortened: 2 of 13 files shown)


src/main/java/org/openmrs/addonindex/domain/artifactory/ArtifactoryDownloadCounts.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents the complete list of add-ons that we want to index the versions of.
+ * The source of this data is a static file on github (which module authors update via pull requests).
+ */
+public class ArtifactoryDownloadCounts {
```
djazayeri: Wouldn't this be BintrayDownloadCounts? Why artifactory?

src/test/java/org/openmrs/addonindex/backend/BintrayTest.java
```diff
@@ -58,7 +59,10 @@ public void testHandleJson() throws Exception {
 		addOnToIndex.setBackend(Bintray.class);
 		addOnToIndex.setBintrayPackageDetails(new BintrayPackageDetails("openmrs", "owa",
 				"openmrs-owa-conceptdictionary"));
-		
+
+		when(restTemplate.getForObject("https://bintray.com/statistics/packageGeoStats?&pkgPath=/openmrs/owa/openmrs-owa-conceptdictionary", JsonNode.class))
+				.thenReturn(new ObjectMapper().readValue(getFileAsString("bintray-download-counts.json"), JsonNode.class));
+
 		AddOnInfoAndVersions info = bintray.handlePackageJson(addOnToIndex, json);
```
djazayeri: you also need to assert something somewhere in this test, right?

dkayiwa:
@reubenvarghese1 are you still working on this?

reubenvarghese1:
@dkayiwa Yes, I am still working on it. It's a small fix and I'll upload a patch by this weekend. Was caught up with college and GCI :)

src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
```diff
@@ -30,5 +30,6 @@
 	 * @throws Exception
 	 */
 	AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) throws Exception;
+	void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception;
```
djazayeri: Needs javadoc to explain what an implementation of this interface should do.

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -65,7 +69,36 @@ public AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) thr
...
+
+	@Override
+	public void fetchDownloadCounts(AddOnToIndex toIndex, AddOnInfoAndVersions infoAndVersions) throws Exception {
+		//We won't be checking for existing info as download counts is dynamic
+		// and has to be fetched each time
+		logger.info("Fetching Download Counts for " + toIndex.getUid());
+		BintrayDownloadCounts downloadCounts;
+		String json;
+		BintrayPackageDetails bintrayPackageDetails = toIndex.getBintrayPackageDetails();
+		String baseurl = "https://bintray.com/statistics/packageGeoStats?&pkgPath=/" + bintrayPackageDetails.getOwner()
```
djazayeri: Why is this "baseurl"? Base implies that you're going to append something to it. Call it downloadStatsUrl or something.

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -65,7 +69,36 @@ public AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) thr
...
+					totalDownloadCounts += downloadCount;
+				}
+				catch (Exception e) {
+					logger.error("Error getting details for " + toIndex.getUid(), e);
+				}
+			}
+			infoAndVersions.setDownloadCounts(totalDownloadCounts);
+		}
+		catch (Exception ex) {
+			throw new RuntimeException("Download counts from " + baseurl + " could not be parsed", ex);
```
djazayeri: Never catch a checked exception and rethrow it as an unchecked RuntimeException. Yes, this seems convenient to do sometimes, but it's never correct.
Also in this specific case our method is _allowed_ to throw Exception, so you can throw this as a plain Exception, or just let it bubble up.

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -98,8 +131,21 @@ AddOnInfoAndVersions handlePackageJson(AddOnToIndex addOnToIndex, String package
...
+
+		String downloadCountUrl = downloadCountUrlFor(addOnToIndex);
+		JsonNode arr = restTemplateBuilder.build().getForObject(downloadCountUrl, JsonNode.class);
+		int totalDownloadCount = handleBintrayGeoStats(arr);
+		info.setDownloadCounts(totalDownloadCount);
 		return info;
 	}
+
+	private int handleBintrayGeoStats(JsonNode jsonNode) {
+		int totalDownloadCount = 0;
```
djazayeri: I do see you've done this twice in two ways. I like the other way (using a POJO) because it's self-documenting, and it makes it clear what the response from that URL is going to be.

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -133,4 +179,12 @@ private String packageUrlFor(AddOnToIndex addOnToIndex) {
 				details.getRepo(),
 				details.getPackageName());
 	}
+
+	private String downloadCountUrlFor(AddOnToIndex addOnToIndex) {
```
djazayeri: It's nice to have this method (and you should incorporate it in the other newer code).

src/main/java/org/openmrs/addonindex/backend/BackendHandler.java
djazayeri: Also, now that I'm realizing that the other tools don't handle this, I suggest doing it in a slightly different way.
Introduce new subinterface like

    public interface SupportsDownloadCounts extends BackendHandler

Then you only implement this for Bintray and don't have to have an empty method in the other classes.

src/main/java/org/openmrs/addonindex/domain/AddOnInfoAndVersions.java
```diff
@@ -54,6 +54,8 @@
 	private String hostedUrl;
 	
 	private List<AddOnVersion> versions = new ArrayList<>();
+
+	private Integer downloadCounts;
```
djazayeri: Need to be clearer what this is, e.g. "downloadCountInLast30Days". (I forget exactly what it is.)

src/main/java/org/openmrs/addonindex/domain/BintrayDownloadCounts.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Defines POJO for parse the Download counts of a module
+ *
```
djazayeri: would be nice to paste an example of the JSON here

src/main/java/org/openmrs/addonindex/scheduled/FetchDetailsToIndex.java
```diff
@@ -109,6 +109,7 @@ void getDetailsAndIndex(AddOnToIndex toIndex) throws Exception {
 			}
 			
 			if (fetchExtraDetails) {
+				handler.fetchDownloadCounts(toIndex, infoAndVersions);
```
djazayeri: assuming you've added a new subinterface like I suggest above, this would become:

    if (handler instanceof SupportsDownloadCounts) {
        hander.fetchDownloadCounts(...);
    }

src/main/ui/app/route/Show.jsx
```diff
@@ -203,6 +203,10 @@ export default class Show extends Component {
                                     })}
                                     </h5></td>
                                 </tr>
+                                <tr>
+                                    <th>Total downloads</th>
```
djazayeri: Don't we only know downloads in the last 30 days, or something like that?

src/main/ui/app/route/Show.jsx
```diff
@@ -203,6 +203,10 @@ export default class Show extends Component {
...
+                                    <td>{addon.downloadCounts ? addon.downloadCounts : 0}</td>
```
djazayeri: In the case where we don't have this, I would display a ?, not 0.

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -100,6 +125,19 @@ AddOnInfoAndVersions handlePackageJson(AddOnToIndex addOnToIndex, String package
...
 		return info;
 	}
+
+	private Integer handleBintrayGeoStats(AddOnToIndex toIndex, BintrayDownloadCounts downloadCounts) {
+		int totalDownloadCount = 0;
+		for (Integer downloadCount : downloadCounts.getTotalDownloads().values()) {
+			try {
+				totalDownloadCount += downloadCount;
+			}
+			catch (Exception e) {
```
djazayeri: the code in this block cannot throw an exception. Get rid of this try/catch

src/main/java/org/openmrs/addonindex/backend/Bintray.java
```diff
@@ -65,7 +69,28 @@ public AddOnInfoAndVersions getInfoAndVersionsFor(AddOnToIndex addOnToIndex) thr
...
+		//We won't be checking for existing info as download counts is dynamic
+		// and has to be fetched each time
+		logger.info("Fetching Download Counts for " + toIndex.getUid());
+		BintrayDownloadCounts downloadCounts;
+		Integer totalDownloadCounts = 0;
+		String json;
+		String downloadStatsUrl = downloadCountUrlFor(toIndex);
+		try {
+			json = restTemplateBuilder.build().getForObject(downloadStatsUrl, String.class);
+			downloadCounts = mapper.readValue(json, BintrayDownloadCounts.class);
```
djazayeri: looking at this code again I see that this class should be called BintrayGeoStats, not BintrayDownloadCounts. Can you do a quick refactor->rename?

src/main/java/org/openmrs/addonindex/backend/SupportsDownloadCounts.java
```diff
@@ -0,0 +1,22 @@
+package org.openmrs.addonindex.backend;
+
+import org.openmrs.addonindex.domain.AddOnInfoAndVersions;
+import org.openmrs.addonindex.domain.AddOnToIndex;
+
+/**
+ *Represents a Strategy for fetching number of downloads over the past 30 days.
```
djazayeri: Actually, in this comment and in the method-level comment, please don't be so specific about the "in last 30 days". The idea would be that various back-ends might allow us to fetch different kinds of download counts, and we'll have to be flexible about it.
So that means that when we _store_ a download count we need to be precise about what kind (which is why I said it's important to say "in last 30 days" in the field name that we're storing), but the code should expect that we'll eventually need to support other things, hence this method should just say "download counts"

src/main/ui/app/route/Show.jsx
```diff
@@ -203,6 +203,10 @@ export default class Show extends Component {
                                     })}
                                     </h5></td>
                                 </tr>
+                                <tr>
+                                    <th>Downloads in the last 30 days</th>
+                                    <td>{addon.downloadCounts ? addon.downloadCounts : "?"}</td>
```
djazayeri: Why is this field name just "downloadCounts" instead of in 30 days?

src/test/java/org/openmrs/addonindex/backend/BintrayTest.java
```diff
@@ -23,11 +14,19 @@
...
+import static org.hamcrest.beans.HasPropertyWithValue.hasProperty;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.when;
+import static org.openmrs.addonindex.TestUtil.getFileAsString;
 
 public class BintrayTest {
-	
+
```
djazayeri: This class has no real changes in it. Did you accidentally remove the test?

PR #37, opened 2018-01-11
AO-15 Add google analytics to Add-Ons

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/AO-15

### DIFF
--- a/src/main/ui/app/index.jsx
+++ b/src/main/ui/app/index.jsx
@@ -12,9 +12,21 @@ import React from "react";
 import {render} from "react-dom";
 import {Router, hashHistory} from "react-router";
 import routes from "./routes";
+import ReactGA from 'react-ga';
+
 import "./sass/addonindex.scss";
 
+if (GA_ID) {
+    ReactGA.initialize(GA_ID);
+}
+
+function trackView() {
+    if (GA_ID) {
+        ReactGA.pageview(window.location.pathname + window.location.search);
+    }
+}
+
 render(
-        <Router history={hashHistory} routes={routes}/>,
+        <Router onUpdate={trackView} history={hashHistory} routes={routes}/>,
         document.getElementById('root')
-)
+);
--- a/src/main/ui/webpack.config.js
+++ b/src/main/ui/webpack.config.js
@@ -10,6 +10,7 @@
 
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
+const webpack = require('webpack');
 
 module.exports = function (env) {
     const BUNDLE_FILENAME = env === "prod" ? "app/bundle.min.js" : "app/bundle.js";
@@ -46,6 +47,9 @@ module.exports = function (env) {
             new ExtractTextPlugin(CSS_FILENAME),
             new HtmlWebpackPlugin({
                 template: 'index.ejs'
+            }),
+            new webpack.DefinePlugin({
+                GA_ID: env === 'prod' ? JSON.stringify('UA-16695719-3') : undefined
             })
         ],
         resolve: {
--- a/src/main/java/org/openmrs/addonindex/backend/Bintray.java
+++ b/src/main/java/org/openmrs/addonindex/backend/Bintray.java
@@ -16,6 +16,7 @@ import java.time.OffsetDateTime;
 import org.openmrs.addonindex.domain.AddOnInfoAndVersions;
 import org.openmrs.addonindex.domain.AddOnToIndex;
 import org.openmrs.addonindex.domain.AddOnVersion;
+import org.openmrs.addonindex.domain.BintrayGeoStats;
 import org.openmrs.addonindex.util.Version;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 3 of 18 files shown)


src/main/ui/app/index.jsx
```diff
@@ -13,8 +13,15 @@ import {render} from "react-dom";
 import {Router, hashHistory} from "react-router";
 import routes from "./routes";
 import "./sass/addonindex.scss";
+import ReactGA from 'react-ga';
+
+ReactGA.initialize('UA-16695719-3');
```
djazayeri: This token needs to be externalized into the configuration file.

src/main/ui/app/index.jsx
```diff
@@ -13,8 +13,15 @@ import {render} from "react-dom";
...
+
+function trackView() {
+    ReactGA.pageview(window.location.pathname + window.location.search);
```
djazayeri: We should only log these page views for the prod server (i.e. if it's not configured, then don't log anything)

BartlomiejRasztabiga: How can we know on which environment are we running the app?

djazayeri: You don't actually need to know that. Basically, you only want to log things if a (to be created) config setting like googleanalytics.url is set

BartlomiejRasztabiga: In which file should I place this property?

BartlomiejRasztabiga:
@djazayeri I've made requested changes

src/main/ui/app/index.jsx
reubenvarghese1: @BartlomiejRasztabiga I can see that you haven't defined the variable in the right place i.e. the "external" key within which you defined the variable is to be used only for listing the external libraries that would be required as mentioned in the comment there. You need to define this variable below this https://github.com/openmrs/openmrs-contrib-addonindex/blob/02f924dd370071b4e65edf8595bae4d4af2e33bf/src/main/ui/webpack.config.js#L12

src/main/ui/webpack.config.js
```diff
@@ -54,7 +54,12 @@ module.exports = function (env) {
         externals: { // JS libraries that we link from a CDN in index.ejs
             'react': 'React',
             'react-dom': 'ReactDOM',
-            'react-router': 'ReactRouter'
+            'react-router': 'ReactRouter',
+            'Config': JSON.stringify(env === "prod" ? {
```
reubenvarghese1: @BartlomiejRasztabiga I can see that you haven't defined the variable in the right place i.e. the "external" key within which you defined the variable is to be used only for listing the external libraries that would be required as mentioned in the comment there. You need to define this variable below the 16th line.

BartlomiejRasztabiga: So when I got this const at 16th line, how can I access it from index.jsx? I can't find that on the Internet.

BartlomiejRasztabiga:
@djazayeri @reubenvarghese1 How can I access const defined in webpack config?

reubenvarghese1:
Check  [this](https://stackoverflow.com/questions/46555505/setting-a-config-value-in-webpack-config-and-access-it-from-react-component) out. Not sure whether it's what would be the right thing to do in this case. I'd wait for @djazayeri's take on this.

BartlomiejRasztabiga:
@djazayeri @reubenvarghese1 I've made requested changes :)

src/main/ui/webpack.config.js
```diff
@@ -46,6 +46,9 @@ module.exports = function (env) {
             new ExtractTextPlugin(CSS_FILENAME),
             new HtmlWebpackPlugin({
                 template: 'index.ejs'
+            }),
+            new webpack.DefinePlugin({
+                GA_ID: env === 'prod' ? JSON.stringify('UA-16695719-3') : undefined
```
djazayeri: This is not right. This is supposed to be a secret, and it should not live in the code at all, but in an external config file (which is set up for the deployment).
I would expect a line in the README file to explain that this can be set.

(Sorry for the delay I had missed the email notifications on this so I just saw it when looking at another PR.)

dkayiwa:
@BartlomiejRasztabiga are you still interested in finishing this up?

BartlomiejRasztabiga:
@dkayiwa Yes, I will finish this possibly tomorrow. I've just forgot about this PR, sorry :)

dkayiwa:
@BartlomiejRasztabiga did you remember to finish this? 😊

src/main/ui/webpack.config.js
BartlomiejRasztabiga: @djazayeri Is these certain file which includes env variables or should I create my own?

dkayiwa: @reubenvarghese1 do you have any response to the above?

dkayiwa: @BartlomiejRasztabiga are you now in position to continue? 😊

reubenvarghese1: @BartlomiejRasztabiga You need not worry about this issue if you are busy. We just need to know if you will be working on this. If not, we will make it ready for work so that someone in the future may work on it. This helps us prevent stagnant issues :-)

EtherDEXchange: I want to complete this but first I need to know if there currently is some way of setting env variables during deployment or can I implement my system, e.g. using method @reubenvarghese1 provided in the comment above.

BartlomiejRasztabiga: I want to complete this but first I need to know if there currently is some way of setting env variables during deployment or can I implement my system, e.g. using method @reubenvarghese1 provided in the comment above.

reubenvarghese1: @BartlomiejRasztabiga Makes sense. Since I am not aware of what happens during Deployment, I would wait for @djazayeri to clear your doubt .

djazayeri: Hmm, looking at this again I realize a few things:
1. we don't have a code example of this yet. (all configurable things so far are on the server-side, not the client code)
2. this id isn't _super_ secret because a determined attacker could probably get it from the JS.

So I can live with you doing it the way you've done here.

BartlomiejRasztabiga:
Sorry, I've messed something up with GitHub here. I've created a new PR. :)
https://github.com/openmrs/openmrs-contrib-addonindex/pull/49

PR #44, opened 2018-06-04
Add OpenMRS Sync 2.0 to indexing

tomaszmueller:
This commit is to make AddOns index the OpenMRS Sync 2.0 module

### DIFF
--- a/src/main/resources/add-ons-to-index.json
+++ b/src/main/resources/add-ons-to-index.json
@@ -105,26 +105,6 @@
       }
     },
     {
-    "uid": "org.openmrs.module.sync2",
-    "type": "OMOD",
-    "name": "Sync 2.0",
-    "tags": [
-        "sync"
-    ],
-    "maintainers": [
-      {
-        "name": "Tomasz Mueller",
-        "url": "https://github.com/tomaszmueller"
-      }
-    ],
-    "backend": "org.openmrs.addonindex.backend.Bintray",
-    "bintrayPackageDetails": {
-      "owner": "tomaszmueller",
-      "repo": "omod",
-      "package": "openmrs-module-sync2"
-      }
-    },
-    {
       "uid": "org.openmrs.module.accesslogging",
       "status": "DEPRECATED",
       "type": "OMOD",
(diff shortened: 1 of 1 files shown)


src/main/resources/add-ons-to-index.json
```diff
@@ -104,6 +104,26 @@
...
+    {
+    "uid": "org.openmrs.module.sync2",
+    "type": "OMOD",
+    "name": "Sync 2.0",
+    "tags": [
+        "sync"
+    ],
+    "maintainers": [
+      {
+        "name": "Tomasz Mueller",
```
dkayiwa: Shouldn't this be? https://github.com/openmrs/openmrs-module-sync2

reubenvarghese1:
Hi @tomaszmueller Thanks for the PR. Since your module is an OMOD, can you please move your code such that the module is alphabetically in the right place  with the other omods?

tomaszmueller:
@reubenvarghese1 Done

src/main/resources/add-ons-to-index.json
```diff
@@ -3233,6 +3213,23 @@
...
+    "name": "Sync 2.0",
+    "maintainers": [
+      {
+        "name": "Tomasz Mueller",
+        "url": "https://github.com/tomaszmueller"
+      }
+    ],
+    "backend": "org.openmrs.addonindex.backend.Bintray",
+    "bintrayPackageDetails": {
+      "owner": "tomaszmueller",
```
reubenvarghese1: Hi @tomaszmueller , on further inspection, I realised that the Bintray details may not be accurate. Could you re-check that for me please? 
I believe that https://bintray.com/openmrs/omod/sync2 is the location of the module?
If yes, then the owner is openmrs

src/main/resources/add-ons-to-index.json
```diff
@@ -3233,6 +3213,23 @@
...
+      "repo": "omod",
+      "package": "openmrs-module-sync2"
```
reubenvarghese1: Similarly, the package is sync2

tomaszmueller:
OK, @reubenvarghese1 changes applied. Anything else?

reubenvarghese1:
@tomaszmueller  Looks good to me, thanks!

djazayeri:
@tomaszmueller this isn't exactly the right place for this conversation, but can I suggest that you should change the naming you're using here to just "Sync 2" and not "Sync 2.0"? Because there will be multiple releases, 2.0, 2.1, 2.2, etc. And it would be pretty confusing to say "Sync 2.0, version 2.1".

PR #49, opened 2018-07-19
AO-15 Add google analytics to Add-Ons

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/AO-15


reubenvarghese1:
We've tested it out and it works great so far. Thanks @BartlomiejRasztabiga

PR #50, opened 2018-07-23
 Fixed broken react-select by using v1.2.1

reubenvarghese1:
Fixed broken react-select by using v1.2.1

### DIFF
--- a/src/main/ui/package.json
+++ b/src/main/ui/package.json
@@ -24,7 +24,7 @@
     "react-bootstrap": "^0.30.7",
     "react-dom": "^15.4.2",
     "react-router": "^3.0.2",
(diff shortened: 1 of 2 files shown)


src/main/ui/package.json
```diff
@@ -24,7 +24,7 @@
     "react-bootstrap": "^0.30.7",
     "react-dom": "^15.4.2",
     "react-router": "^3.0.2",
-    "react-select": "^1.0.0-rc.5"
+    "react-select": "1.2.1"
```
djazayeri: This should be "1.x" (if 1.3.0 is acceptable) or "1.2.x". I.e. we'd want to pull in maintenance releases, and we're just trying to avoid 2.x.

PR #51, opened 2018-07-24
 Made modifications to use react-select v1.x

reubenvarghese1:
Made modifications to use react-select v1.x


reubenvarghese1:
@djazayeri I guess this commit should address your comments on the previous PR :-)

PR #54, opened 2018-08-25
Added more info to the developer and the LBAC, SysAdmin modules

suthagar23:
I have just added some more information for the existing modules which are managed by myself.


reubenvarghese1:
Thanks @suthagar23 , merged!

PR #70, opened 2019-06-14
AO-22 Add support for Advanced Query Syntax

reubenvarghese1:


reubenvarghese1:
@djazayeri @Ruhanga please find an initial commit towards implementing AO 22. Do let me know of any changes that you find may be necessary(including the approach itself).

djazayeri:
High-level comment: avoid hand-coding each individual advanced parameter. I would expect that most of the supported things map directly to one field in the elasticsearch index, so I would define a js object giving these mappings, and have some code that finds things before the : and checks if they're keys on that object.

djazayeri:
Also, I hope we can support some common things like "exact phrase" or -exclude. Can you find an existing advanced query parsing library we can use? A quick google search came up with this: https://github.com/nepsilon/search-query-parser (though there may be better ones).

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -20,68 +21,50 @@ class SearchBox extends Component {
...
-                let querySplit = advancedQuery.split(' ');
-                let tempQuery = "";
-                querySplit.forEach(m => {
-                    if (m.includes(":")){
-                        let [key, value] = m.split(':');
-                        if (key === "type"){
-                            queryComponents["type"] = value.toUpperCase();
+            let url = "/search?";
+            // Check if query is an advanced query. We want to use the Parser only if the query is advanced
+            if (advancedQuery.includes(":") || advancedQuery.includes("-")){
```
djazayeri: I think we want a way for the user to include a : or a - in the query. Can we document what it is in the help text? (Maybe not even so important to support this.)

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -31,10 +31,16 @@ export default class SearchPage extends Component {
...
+        const advancedQuery = this.props.location.query;
         const query = this.props.location.query.q;
         const type = this.props.location.query.type;
-	const tag = this.props.location.query.tag;
-	const searchKey = `type:${type ? type : "all"} query:${query ? query : ""} tag:${tag ? tag : ""}`;
+        const tag = this.props.location.query.tag;
+        const exclude = this.props.location.query.exclude;
+        const moduleid = this.props.location.query.moduleid;
+        const name = this.props.location.query.name;
+        const status = this.props.location.query.status;
```
djazayeri: Why do we have to enumerate all of these here? Can't they be encapsulated within SearchBox? E.g. this class only needs to care about a single search string ("query"), right?

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
```diff
@@ -12,10 +12,7 @@
 
 import java.util.Collection;
 
-import org.openmrs.addonindex.domain.AddOnInfoAndVersions;
-import org.openmrs.addonindex.domain.AddOnInfoSummary;
-import org.openmrs.addonindex.domain.AddOnType;
-import org.openmrs.addonindex.domain.AddOnVersion;
+import org.openmrs.addonindex.domain.*;
```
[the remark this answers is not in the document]
reubenvarghese1: Sorry about that, I've now disabled it in my IDE settings

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -20,68 +21,50 @@ class SearchBox extends Component {
...
+            query: query
         });
     }
 
-    parseQuery(){
-        if (this.state.query){
-            let advancedQuery = this.state.query;
+    doSearch() {
+        if (this.state.query) {
+            let advancedQuery = this.state.query.toLowerCase();
```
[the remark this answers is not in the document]
reubenvarghese1: Typically, lowercasing would mess it up as the keyword field is case sensitive. However, fortunately, only our name field is having a mix of case. The rest of the fields are all lowercased. Even while performing exact matching on the name field, I reckon that both "Observations Administration" and "observations administration" should produce the same result. I have ensured this by adding a normalizer to the name field. This normalizer basically helps make the keyword field case-insensitive

src/main/ui/app/component/SearchBox.jsx
reubenvarghese1: So do you mean to say that every query should have an advanced query like structure? IMO, certain users just prefer searching for text instead of having to worry about the structure of the advanced query and I believe we should continue to support that. Those who wish to go advanced will also have the option to do so. However, should you feel that we should do away with a simple query, do let me know and I will implement it :)

src/main/ui/app/route/SearchPage.jsx
reubenvarghese1: Sorry you had to see that. It was very foolish of me to do so :)

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -21,43 +22,57 @@ class SearchBox extends Component {
...
+        if (this.state.query) {
+            let query = this.state.query.toLowerCase();
+            const options = {keywords: ['uid', 'type', 'tag', 'query', 'moduleid', 'status', 'name'], offsets: false};
+            //Basic Regex Matching to fix query inconsistencies
+            //Removing all extra spaces i.e. two or more
+            query = query.replace(/\s+/g,' ').trim();
+            //Removing all spaces to the left of search keys
+            query = query.replace(new RegExp("\\s+:","g"),":");
+            //Removing all spaces to the right of search keys
+            query = query.replace(new RegExp(":\\s+","g"),":");
```
djazayeri: Can you pull all this into a helper function like normalizeQuery()?

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -21,43 +22,57 @@ class SearchBox extends Component {
...
             let url = "/search?";
-            if (this.state.type) {
-                url += "type=" + this.state.type;
+            // Check if query is an advanced query. We want to use the Parser only if the query is advanced
+            if (query.includes(":") || query.includes("-")) {
```
djazayeri: Peeking at https://www.npmjs.com/package/search-query-parser it says

> If no keywords or ranges are specified, or if none are present in the given search query, then searchQuery.parse will return a string if tokenize is false

so you should rely on this, and avoid doing this test yourself. (E.g. what if I query for: notAValidKey:foo ?)

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -21,43 +22,57 @@ class SearchBox extends Component {
...
+                let searchQueryObj = searchQuery.parse(query, options);
+                Object.keys(searchQueryObj).forEach(function (key) {
+                    console.log(key);
```
djazayeri: need to remove this before merging the PR

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
```diff
@@ -38,9 +41,14 @@ public AddOnController(Index index) {
 	
 	@RequestMapping(method = RequestMethod.GET, value = "/api/v1/addon")
 	public Collection<AddOnInfoSummary> search(@RequestParam(value = "type", required = false) AddOnType type,
-	                                           @RequestParam(value = "q", required = false) String query, 
-						   @RequestParam(value = "tag", required = false) String tag) throws Exception {
-		return index.search(type, query, tag);
+											   @RequestParam(value = "q", required = false) String query,
```
djazayeri: In the big picture it would be nice to avoid changing method signature and adding these parameters all over the place, by just doing the query parsing entirely within the ElasticSearchIndex class.

src/main/resources/elasticsearch/addOnInfoAndVersions-settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
djazayeri: Please give some more justification of why you're doing this (ideally in a code comment somewhere)

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -21,43 +22,57 @@ class SearchBox extends Component {
...
-                      });
+            query: query
+        });
     }
 
     doSearch() {
-        if (this.state.type || this.state.query) {
+        if (this.state.query) {
+            let query = this.state.query.toLowerCase();
+            const options = {keywords: ['uid', 'type', 'tag', 'query', 'moduleid', 'status', 'name'], offsets: false};
```
djazayeri: pull this const outside of the function

src/main/ui/app/component/SearchBox.jsx
```diff
@@ -21,43 +22,57 @@ class SearchBox extends Component {
...
+                        }
+                        else if (key === "type"){
+                            url += "&type=" + searchQueryObj[key].toUpperCase();
+                        }
+                        else if (key === "exclude"){
+                            if (searchQueryObj[key].text){
+                                url += "&exclude=" + searchQueryObj[key].text;
+                            }
+                        }
+                        else if (key === "status"){
```
djazayeri: you're handling "type" and "status" in the same way, can you combine these branches?

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -43,15 +41,13 @@ export default class SearchPage extends Component {
...
-        if (query) {
-			url += "&q=" + query;
-		}
-		if (tag) {
-			url += "&tag=" + tag;
-		}
+
+        Object.keys(query).forEach(function (key) {
+            if (query[key]){
+                url += ("&" + key + "=" + query[key]);
```
djazayeri: why are you constructing an url both in SearchBox and here?

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -77,13 +73,15 @@ export default class SearchPage extends Component {
...
+        let query = this.props.location.query;
+
         if (this.state && this.state.error) {
             return <div>{this.state.error}</div>
         }
         else {
             return (
                     <div>
-                        <SearchBox initialQuery={this.props.location.query.q}/>
+                        <SearchBox initialQuery={`${query.type ? "type:" + query.type : ""} ${query.q ? "query:" + query.q : ""} ${query.tag ? "tag:" + query.tag : ""} ${query.exclude ? "-" + query.exclude : ""} ${query.name ? "name:" + query.name : ""} ${query.status ? "status:" + query.status : ""} ${query.moduleid ? "moduleId:" + query.moduleid : ""} ${query.uid ? "uid:" + query.uid : ""}`}/>
```
djazayeri: The fact that you're doing this feels like a code smell.

Maybe this would work better if SearchBox continues to just be mainly a UI element, and you move the parsing into this SearchPage class?

src/main/ui/app/component/SearchBox.jsx
reubenvarghese1: @djazayeri , although from what's mentioned in the docs, it seems that the parser just returns a string, it actually returns something like query: "user query" . This makes it impossible to differentiate a normal query from an advanced one. That is why I had implemented this test to check the pre-parsed query

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
reubenvarghese1: Yes, that's true, I have currently asked around on forums. Let's hope a java based parser that's similar to the one we're currently using shows up

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -98,8 +105,25 @@ public void index(AddOnInfoAndVersions infoAndVersions) throws IOException {
 	}
 	
 	@Override
-	public Collection<AddOnInfoSummary> search(AddOnType type, String query, String tag) throws IOException {
+	public Collection<AddOnInfoSummary> search(AddOnType type, String query, String tag, String uid, String moduleId,
+											   String name, String exclude, AddOnStatus status) throws IOException {
 		BoolQueryBuilder boolQB = QueryBuilders.boolQuery();
+		if (name != null) {
+			//Exact match on name
+			boolQB.filter(QueryBuilders.matchQuery("name.raw", name));
```
djazayeri: this should be case-insensitive.
probably all of them should be. Are they?

reubenvarghese1: Yes, all are case-insensitive. Keywords are by default case-sensitive. To make this field case-insensitive is why I added a [normalizer](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/normalizer.html) to it(this feature is only available starting v5.2 of elasticsearch. So, I guess we should update the ReadME to include this)

src/main/java/org/openmrs/addonindex/rest/AddOnController.java
djazayeri: Conclusion: we didn't find a Java library for this, so we'll proceed doing it in the front end in JS.

src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
djazayeri: Please update the README. Is that about also sourcing the -setting.json file, or do we also need to update the deployment scripts to use a later version of elasticsearch?

src/main/ui/app/route/SearchPage.jsx
```diff
@@ -30,28 +35,56 @@ export default class SearchPage extends Component {
...
+            });
+        }
+        else {
+            url += "&q=" + query;
         }
-        if (query) {
-			url += "&q=" + query;
-		}
-		if (tag) {
-			url += "&tag=" + tag;
```
djazayeri: This change implies there should be a change to the URL of the link for searching by tag, right? I don't see it changed in this PR.

src/main/resources/elasticsearch/addOnInfoAndVersions-settings.json
djazayeri: Resolved

PR #72, opened 2019-06-15
AO 23 Fix cache issue in AddOns

reubenvarghese1:
List of changes in this commit:

- **Upgraded Webpack to V4 from V2**

- **Fixed cache issue using contentcache(supported from webpack v4.3 onwards) to generate a hash for file names**

- **Optimized output bundle.js by using splitting chunks**

To ensure smooth migration it is recommended that **npm version is >6** and  **nodejs >10**
Reference: https://webpack.js.org/guides/caching

[Issue: https://issues.openmrs.org/browse/AO-23](https://issues.openmrs.org/browse/AO-23)

EDIT: It looks like CI passes without any issues even with NodeJs 8 installed(but we should update it to use a higher version I guess?)

### DIFF
--- a/src/main/ui/.babelrc
+++ b/src/main/ui/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    "es2015",
-    "react"
+    "@babel/env",
+    "@babel/preset-react"
   ]
 }
--- a/src/main/ui/package-lock.json
+++ b/src/main/ui/package-lock.json
@@ -13,6 +13,400 @@
         "@babel/highlight": "7.0.0-rc.2"
       }
     },
+    "@babel/core": {
+      "version": "7.4.5",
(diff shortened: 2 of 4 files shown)


djazayeri:
I haven't ever used the latest webpack, so I'll trust our tests that you've done this right.

dkayiwa:
Does any one know the cause of this CI failure? https://ci.openmrs.org/browse/AD-BAOIS-106

reubenvarghese1:
@dkayiwa , thanks for checking this out. I am aware of the failure and I am currently working with Cintia to get this running. Basically, we need to delete the existing node_modules folder in the server and use a higher version of nodejs to build AddOns. The failure will be fixed by this weekend so kindly ignore this during the scrum meetings until this weekend :)

PR #75, opened 2019-07-23
AO-25 Recent Releases page

reubenvarghese1:
This PR contains code and its corresponding tests for AO 25.

Issue: https://issues.openmrs.org/browse/AO-25

### DIFF
--- a/src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
+++ b/src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
@@ -23,6 +23,8 @@ import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.BoostingQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
+import org.elasticsearch.search.sort.SortBuilder;
+import org.elasticsearch.search.sort.SortBuilders;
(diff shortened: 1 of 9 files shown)


src/main/java/org/openmrs/addonindex/service/ElasticSearchIndex.java
```diff
@@ -191,6 +193,24 @@ public AddOnInfoAndVersions getByModulePackage(String modulePackage) throws IOEx
 				.map(sr -> sr.source)
 				.collect(Collectors.toList());
 	}
+
+	@Override
+	public Collection<AddOnInfoAndVersions> getRecentReleases(int resultSize) throws IOException {
```
reubenvarghese1: Returns a collection of type AddOnInfoAndVersions instead of AddOnInfoSummaryAndStats because I have displayed the Release date on the release page in the UI. The release date is not available in AddOnInfoSummaryAndStats. Do let me know if the release date is not needed in that page and I can remove it(my thinking was that the release page should have two key pieces of info for the user to be able to figure out if there has been a newer version of the module that they use been released. Either the module version or the release date can be indicative of a newer version. For ex: the user doesn't remember the exact module version but possibly remembers that the last time that they updated that module was sometime around last time but now they see that a new version had been released 4 days back.

reubenvarghese1:
Perfect, that works!

dkayiwa:
@reubenvarghese1 did you see these merge conflicts?

